PR #19, opened 2021-02-04
Validating filetypes

cjrace:
Adding validation to prevent non-csv file types from being able to be screened. 

This is over and above the 'suggestion' made to browsers by `fileInput(..., accept = ".csv")` in the ui.R script which only tells a browser to expect a certain filetype, and doesn't prevent this from being overridden. As such, users can manually select all files in their file browser window and then upload and attempt to screen invalid file types, causing the app to crash without any feedback to the user as to why. These changes should stop this from happening.

This seems to wipe out our css formatting of the colour for upload bars, though that's not a breaking issue.


cjrace:
Have managed to fix the wiping out of the css on the upload bars. UI tests using shinytests weren't really doing much and were causing pain in the GHA workflow so I've dropped them entirely. Will consider RSelenium instead

cjrace:
The css change was impossible to spot unless you rendered the entire file. So I've ran it through a beautifier (https://www.cleancss.com/css-beautify/) so that when lines are edited we can track them better. For reference my change was to add `!important` into the argument I wanted to make override other ones `.progress-bar(background-color)`

PR #20, opened 2021-02-05
Adding validation on file upload, package updates, css cleaning

cjrace:
1. Adding validation to prevent non-csv file types from being able to be screened.

> This is over and above the 'suggestion' made to browsers by fileInput(..., accept = ".csv") in the ui.R script which only tells a browser to expect a certain filetype, and doesn't prevent this from being overridden. As such, users could previously manually select `all files` in their file browser window and then upload and attempt to screen invalid file types, causing the app to crash without any feedback to the user as to why. 
>
>This PR should stop this from happening by giving a friendly error with useful feedback if an invalid filetype (anything other than csv) is uploaded. It.
> 
> The screening button should not appear while a warning is present, and warnings should dissapear if a correct filetype is uploaded.

2. The CSS has been updated to force the colour of the upload bars.

> I have managed to fix the wiping out of the css on the upload bars. UI tests using shinytests weren't really doing much and were causing pain in the GHA workflow so I've dropped them entirely. Will consider RSelenium instead in a seperate branch as this should be stable enough for the time being.
>
>For reference, my change was to add `!important` into the argument I wanted to make override other ones (`.progress-bar(...; background-color: ...)`).

3. I have 'beautified' the css code so it should be easier to track changes in the future.
 
> The css change was impossible to spot unless you rendered the entire file. So I've ran it through a beautifier (https://www.cleancss.com/css-beautify/) so that when lines are edited we can track them better. 

4. `renv::update()` was ran, updating a number of packages including shiny (to 1.6.0) and renv itself.


cjrace:
Have reverted the package updates for now as they seem to cause an issue with the deployment (have posted in the R Shiny Developers teams channel). So this PR is now for 1,2, and 3 only

lauraselby (approved):

PR #27, opened 2021-03-17
Allowing LA data in LAD rows

cjrace:
This should allow LA columns to be completed in LAD rows without the overcomplete test flagging it as a fail.

This also updates the message on the indicator_grouping test to clarify that spaces are included.


sarahmwong:
Looks good! Nick's file now passes :) I've had a play with removing higher geographies at the LAD geographic_level and it comes up with a message for when region isn't there, but not for when LA isn't there - is that something we should add in? The recommendation message at the moment for region says:

 "Both region_code and region_name have missing values for local authority district rows in the data file. 
- It is recommended to include the information from these columns for local authority district level data."

cjrace:
I think that's okay to leave it as it is @sarahmwong . Unless we would always expect LA information to be completed for LAD rows? If we would expect that information to always be there then I can add a warning in for it

sarahmwong:
Ok fab :) Yep thinking about it, there might be cases where teams are just uploading LAD-level info but don't want the LA hierarchy in the data!

sarahmwong (approved):
Approved - file passes screener as expected

PR #29, opened 2021-05-25
Harmonisation

lauraselby:
This PR extends the existing country_code and region_code checks to check combinations between code and name against a reference table instead of comparing just the code. 

Other changes to note: 

- Added csv files for expected country and region combination lookups. I've also added an LA file, but no check for this yet.
- Updated test data (from failing tests only) to correct Y&TH labelling

I haven't added any new tests to the existing country/region code tests yet as these were already correctly failing on code errors. I could add more to test label errors and code/label combo errors but that felt a bit overkill - happy to do it if that's how thorough we're being elsewhere.


cjrace (changes requested):
Spotted a few things, want me to fix them and then handle the merging/deploying or do you want to fix them @lauraselby ?

1. `expected_country_codes` and `expected_region_codes` can be deleted as I think they're redundant?
2. The region_code check messages should be updated to be region rather than country
3. Have you added a passes everything xlsx file into the shinytests data folder? What does that do?

lauraselby:
Thanks, I've sorted 1 + 2 and for 3 this file was already there, I've just updated the region data within it :)

cjrace (approved):

PR #32, opened 2021-06-03
Fixing filter groups error

cjrace:
Merging in the fix from the MoJ fixing branch around filter groups causing errors if they are read as empty character strings rather than NA.


sarahmwong (approved):
Tested on the MoJ file in dev and works :)

PR #33, opened 2021-06-04
Initial changes from Trello

sarahmwong:
- Preview tables now use datatable and are stylised with the app
- Suppression table has % column based on total number of indicator cells as denominator
- Suppression table has sensible headers
- Suppression table only shows when there are suppression symbols, otherwise shows text to say no cells are suppressed
- Nice label for indicators pulls through in indicator table
- Where filter groups exist, these pull through to the tables of filters
- Spinners now there for all tables


lauraselby (approved):

sarahmwong:
Agreed to merge and fix tests once we merge qa-pages to dev

PR #39, opened 2021-06-17
Qa sarah

sarahmwong:
Small PR to add in test scripts for shinytest with new qa tabs.

Temp fix on reset button to let me write the tests with changing files, but still needs a long-term fix for the reset button

Still looking for way to test tabs appearing as they're shown/hidden with shinyjs calls, so shiny and server tests won't pick it up. 
Adding PR now so it doesn't hold off the reset button work :)


cjrace (approved):

cjrace (approved):
Awesome job on figuring this out and making it work!

PR #40, opened 2021-06-21
Sparklines

lauraselby:
I mentioned we could add sparklines to the indicator summary tables, this PR adds basic ones.

I'm undecided how helpful they are and played around with the colours a lot and whether their axis should start at 0 or not (if you do this it can be really hard to see any relative changes).

image

### DIFF
--- a/global.R
+++ b/global.R
@@ -21,6 +21,7 @@ library(shinyFeedback)
 library(DT)
 library(shinyWidgets)
 library(shinycssloaders)
+library(sparkline)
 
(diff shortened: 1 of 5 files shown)


sarahmwong:
I like the sparklines to show the trends across years, don't think the axes not starting at 0 is a major issue as you're looking at YoY trends at a glance rather than comparing the actual numbers. However can't get it to run on my end with passes.everything, might be something with the merge that has gone strange?:

image

lauraselby:
Interesting, I'll clear out my local version and will see if I can get the same error!

PR #42, opened 2021-07-07
Location tests

sarahmwong:
Adding tests for:
- Checking when lower-level geographies are the same code
-  Checking when lower-level geographies share the same name
**Needed to separate out school-level testing as both URNs and LAESTABs should be unique to each school name/vice versa**
- Checking when schools share the same URN and where URNs share the same school name 
- Checking when schools share the same LAESTAB and where LEASTABs share the same school name
-  Checking that any code ":" has the location name "Not available"
- Checking that any location name "Not available" has the code ":"
-  Checking for leading/trailing whitespace in all filter variables (including location filters)

Adding in data and tests to test_that script for unit testing.


tests/shinytest/UI_tests-expected/002.json
```diff
@@ -4,6 +4,6 @@
...
... levels than their corresponding filter. <\/td> <td> PASS <\/td> <\/tr>\n  <tr> <td> No filter groups are included in the col_name column. <\/td> <td> PASS <\/td> <\/tr>\n  <tr> <td> All of the filter_group values are unique. <\/td> <td> PASS <\/td> <\/tr>\n  <tr> <td> No filter labels contain leading or trailing whitespace. <\/td> <td> PASS <\/td> <\/tr>\n  <tr> <td> No filters have an indicator_grouping value. <\/td> <td> PASS <\/td> <\/tr>\n  <tr> <td> There are no issues when stripping alpha-numeric characters from filter groups. <\/td> <td> PASS <\/td> <\/tr>\n  <tr> <td> There are no issues when stripping alpha-numeric characters and spaces from indicator groups. <\/td> <td> PASS <\/td> <\/tr>\n  <tr> <td> No filters have an indicator_unit value. <\/td> <td> PASS <\/td> <\/tr>\n  <tr> <td> The indicator_unit values are valid. <\/td> <td> PASS <\/td> <\/tr>\n  <tr> <td> No filters have an indicator_dp value. <\/td> <td> PASS <\/td> <\/tr>\n  <tr> <td> The indicator_dp column only contains blanks or positive integer values. <\/td> <td> PASS <\/td> <\/tr>\n  <tr> <td> The indicator_dp column is completed for all indicators. <\/td> <td> PASS <\/td> <\/tr>\n   <\/tbody> <\/table>"
```
cjrace: Could we make the messages that start with 'every' have a capital at the start? The ones with a variable name at the start are fine, think it's just the 'every' ones that stand out

tests/testthat/mainTests/na_geography.csv
```diff
@@ -0,0 +1,18 @@
+time_period,time_identifier,geographic_level,country_code,country_name,local_enterprise_partnership_code,local_enterprise_partnership_name,trust_id,trust_name,establishment_type,establishment_group,Total_st,Total_st_pc,Total_ehc,Total_ehc_pc,Total_all,Total_all_pc
+2021,Calendar year,Local enterprise partnership,E92000001,England,E92000001,England,,,Total,Total,z,z,430697,100,430697,100
```
cjrace: Think it's fine and seems to do the job for now, but we'll probably need to update the names/codes to be actual LEPs and MATs as the harmonisation work progresses

tests/testthat/test-screenFiles.R
```diff
@@ -63,13 +63,13 @@ test_that("multipleFilterGroupStripped", {
 test_that("zLocationCode", {
   screeningOutput <- testOther("../../tests/testthat/otherData/adding_z_locationCode.csv")
 
-  expect_equal(screeningOutput$results %>% filter(result != "PASS") %>% nrow(), 0)
+  expect_equal(screeningOutput$results %>% filter(result == "FAIL") %>% nrow(), 0)
```
cjrace, line 66: Looks and seems to work fine to me - though wondering what caused you to change it from 'not pass' to 'fail'?

R/mainTests.r
```diff
@@ -1153,6 +1160,454 @@ country_code <- function(data) {
   return(output)
 }
 
+
+# school_laestab_duplicates  ----------------------------------------
+# check that there is a 1:1 relationship between school laestab codes and names
+school_laestab_duplicates <- function(data) {
+  if (!"school_name" %in% names(data)) {
```
cjrace: Until we add in the column checks for school level data I think there's opportunity for bugs here? They might have school level data, though not have the right columns, which would contradict the error message. They also might not have school_laestab, and I think that would cause the app to crash as it could be looking for a non-existent column? (Admittedly, I've not actually tested this)

I'd suggest checking if 'School' is in the geographic_levels, and then perhaps a quick check to see if both school_name and school_laestab are present, and failing the test early if not to prevent the app breaking?

R/mainTests.r
```diff
@@ -1153,6 +1160,454 @@ country_code <- function(data) {
...
+    output <- list(
+      "message" = "School-level data is not present in this data file.",
+      "result" = "IGNORE"
+    )
+  } else {
+    multi_count_code <- data %>%
+      select("school_laestab", "school_name") %>%
+      distinct() %>%
+      add_count(school_name, name = "school_name_n") %>%
```
cjrace, line 1191: Nifty! 😃

R/mainTests.r
```diff
@@ -1153,6 +1160,454 @@ country_code <- function(data) {
...
+      filter(school_name_n > 1) %>%
+      mutate(combo = paste0(school_name, " - ", school_name_n, " different laestabs")) %>%
```
cjrace: Could we make it 'X - **has** Y different laestabs'? Think adding the has makes it slightly clearer what we're meaning? This goes for the other similar messages too

R/mainTests.r
```diff
@@ -1153,6 +1160,454 @@ country_code <- function(data) {
...
+      }
+    }
+  }
+  return(output)
+}
+
+# school_urn_duplicates  ----------------------------------------
+# check that there is a 1:1 relationship between school urns and names
+school_urn_duplicates <- function(data) {
+  if (!"school_name" %in% names(data)) {
```
cjrace: Same bug thoughts as above

R/mainTests.r
```diff
@@ -1153,6 +1160,454 @@ country_code <- function(data) {
...
+
+    if (length(multi_count_name) == 0) {
+      output <- list(
+        "message" = "every geography has one code.",
+        "result" = "PASS"
+      )
+    } else {
+      if (length(multi_count_name) == 1) {
+        output <- list(
+          "message" = paste0("The following geography has multiple codes:'", paste0(multi_count_name), "'.
```
cjrace: Missing a space after the colon

R/mainTests.r
```diff
@@ -1153,6 +1160,454 @@ country_code <- function(data) {
...
+      }
+    }
+  }
+  return(output)
+}
+
+# na_geography -------------------------------------
+# checking if location has code of ":", then name is "not available"
+
+geography_name_codes <- geography_matrix[, 2:3] %>% as.character()
```
cjrace: Think you can add `%>% .[!is.na(.)]` to line 1494 and remove the need for line 1496?

R/mainTests.r
```diff
@@ -1153,6 +1160,454 @@ country_code <- function(data) {
...
+      add_count(name, name = "name_n") %>%
+      add_count(code, name = "code_n")
+
+    multi_count_name <- lookup_creator %>%
+      filter(name_n > 1) %>%
+      mutate(combo = paste0(name, " - ", name_n, " different codes")) %>%
+      select(combo) %>%
+      distinct() %>%
+      pull()
+
```
cjrace, line 1422: This section from 1367 through 1391 is really neat, I like it a lot 😃

R/mainTests.r
```diff
@@ -1153,6 +1160,454 @@ country_code <- function(data) {
...
+    }
+  }
+  return(output)
+}
+
+
+# other_geography_duplicates  ----------------------------------------
+# check that there is a 1:1 relationship between geography codes and names
+# ignore school
+lower_level_geog_names <- geography_matrix[c(7:12, 14:16), 2:3] %>% as.character()
```
cjrace, line 1380: Pretty nifty, didn't realise you could use `as.character()` to convert into a vector

R/mainTests.r
```diff
@@ -1153,6 +1160,454 @@ country_code <- function(data) {
...
+  }
+  return(output)
+}
+
+# other_geography_code_duplicates  ----------------------------------------
+# check that there is a 1:1 relationship between geography names and codes
+
+lower_level_geog_names <- geography_matrix[c(7:12, 14:16), 2:3] %>% as.character()
+
+other_geography_code_duplicates <- function(data) {
```
cjrace, line 1455: Is the lookup_creator object the same as the previous test? If so, is it possible to combine this test with that test (other_geography_duplicates)? It would save a little bit of code, though still works fine as separate if it's faffy or not easily done

R/mainTests.r
```diff
@@ -1153,6 +1160,454 @@ country_code <- function(data) {
...
+    )
+  } else {
+    if (nrow(check_nas) == 0) {
+      output <- list(
+        "message" = paste0("All location with a code of ':' have the corresponding name 'Not available'."),
+        "result" = "PASS"
+      )
+    } else {
+      output <- list(
+        "message" = paste0("A location has a code of ':', but does not have the corresponding name 'Not available'."),
```
cjrace: Change to 'At least one location has a code of' to account for single and plural in terms of affected rows?

R/mainTests.r
```diff
@@ -1153,6 +1160,454 @@ country_code <- function(data) {
...
+    }
+  }
+
+  return(output)
+}
+
+# na_geography_code -------------------------------------
+# checking if location has the name "not available" then its code is ":"
+
+geography_name_codes <- geography_matrix[, 2:3] %>% as.character()
```
cjrace: Again think you can remove the na's in this line and ditch line 1555

R/mainTests.r
```diff
@@ -1498,6 +1953,39 @@ filter_group_duplicate <- function(meta) {
...
+# whitespace_filters -------------------------------------
+# check no filter labels have leading or trailing whitespace.
+
+whitespace_filters <- function(data, meta) {
+  filters <- meta %>%
+    filter(col_type == "Filter") %>%
+    pull(col_name)
+
+  test <- data %>%
+    select(all_of(filters), any_of(as.character(geography_matrix[, 2]))) %>%
```
cjrace: I think you're missing some of the geography columns out here? Should it be `geography_matrix[, 2:6]`? Though changing that will introduce NAs, so they might need removing if you do?

cjrace: This might work? 
`select(all_of(filters), any_of(c(as.character(geography_matrix[, 2:6]) %>% .[!is.na(.)]))) %>%`

R/mainTests.r
```diff
@@ -1498,6 +1953,39 @@ filter_group_duplicate <- function(meta) {
...
+    gather(., "filter", "filter_label") %>%
```
cjrace: I think `gather()` is out of date and we should use `pivot_longer()` now? 

Either way, it's simple but I like how nifty this bit of code is, sets it up really nicely to pull out the detail for the error message 😀

R/mainTests.r
```diff
@@ -1498,6 +1953,39 @@ filter_group_duplicate <- function(meta) {
...
+  test_no_whitespace <- test %>% mutate(filter_label = str_trim(filter_label))
+
+  white_spaces <- setdiff(test, test_no_whitespace) %>% pull(filter_label)
+
+  if (length(white_spaces) == 0) {
+    output <- list(
+      "message" = "No filter labels contain leading or trailing whitespace.",
+      "result" = "PASS"
+    )
+  } else {
```
cjrace, line 2009: Could add a singular version for where `length(white_spaces == 1)`?

R/mainTests.r
```diff
@@ -1153,6 +1160,454 @@ country_code <- function(data) {
...
+      if (length(multi_count_code) == 1 & length(multi_count_name) == 0) {
+        output <- list(
+          "message" = paste0("The following school_name has multiple laestabs '", paste0(multi_count_code), "'. 
+                             <br> - Every school name should have only one laestab and vice versa."),
+          "result" = "FAIL"
+        )
+      } else {
+        if (length(multi_count_code) == 1 & length(multi_count_name) == 1) {
+          output <- list(
+            "message" = paste0("The following school_name has multiple laestabs '", paste0(multi_count_code), "',
```
cjrace: I think these messages where you have words inside the ' ' we could drop the quote marks? They're useful to show specific variable values, but once we're adding text in there too I think they get a little confusing and clutter it a bit. Could add them in higher up when making the combo column, or drop them entirely.

Same then goes across the rest of the similar error messages

cjrace (changes requested):
Looks very good, some great code and nifty solutions in here! I've left a few comments which probably span the full range from helpful to pedantic 😅 let me know if you have any questions about any of them.

tests/testthat/test-screenFiles.R
sarahmwong: Adding in the school/lower level geog tests gave four "Ignore" outputs, which meant the old test to check "not pass" failed :)

R/mainTests.r
sarahmwong: so much neater thanks!

tests/shinytest/UI_tests-expected/002.json
sarahmwong: Yes good point! Amended in the latest commit

tests/testthat/mainTests/na_geography.csv
sarahmwong: Also amended in this commit to use real LEPs and MATs :)

R/mainTests.r
sarahmwong: Good point! Just added to the tests now - had a play around with removing URN and LAESTAB cols and the app displays a fail for the test instead of crashing now

R/mainTests.r
sarahmwong: I had it together originally, but the outputs looked quite untidy if there were multiple codes with multiple names/vice versa! Think splitting it out made it easier to decode the fail message

R/mainTests.r
sarahmwong: Thank you - this worked! Amended now :)

R/mainTests.r
sarahmwong: Moved over to pivot_now - much neater as it removes the NAs from the values! Hadn't seen this one before - thank you!

R/mainTests.r
sarahmwong: Yes dropped these now - agree much neater this way

R/mainTests.r
sarahmwong: Yes added now :)

R/mainTests.r
cjrace: Nice! I'm still not sure about that the initial message saying there isn't any school level data, as that's only going by the presence of the school_name, would probably be more reliable to use `!"School" %in% unique(data$geographic_level)` as the first condition for ignoring, and then if school_name isn't there it would be a fail, like you've done for school_laestab (and school_urn)?

sarahmwong: Ah yes this makes much more sense thank you! Amended in the latest commit :)

PR #43, opened 2021-07-13
Adding warning modal to shinyapps.io version

cjrace:
### What is the problem to solve?

We wanted to add a confirmation modal to ensure that users on the externally hosted version understood the data security risks involved. It's worth double checking my logic and testing it to see if there's bugs/scenarios I've not covered.

The app should function as before for screening files, including after resetting. 

Environment for testing - http://department-for-education.shinyapps.io/dfe-published-data-qa (Laura/Sarah - I've invited your DfE accounts as access to this is still locked down for now)

You can also test this locally, running it as is will recreate the behaviour on rsconnect, and changing the reactiveValue at the top of the server file manually to "shinyapps" will recreate the behaviour on shinyapps.io.

---

### How I've gone about this

To do this, I have added a config file to identify the environment we're in. shinyapps.io gives a default value that we can pull from (rsconnect apparently does too though it seemed unresponsive).

I've then set this environment variable into a reactiveValue, when this is "shinyapps" a different version of the screening button appears, this different button triggers the modal to appear. The old button still works as before if the environment is not shinyapps.io.

To get the confirmation from the modal, I then update another reactiveValue using the callback.

I've updated the observeEvent() trigger to be on either the original screening button, or on the reactiveValue set by the callback from the modal. 

To keep the original screening button behaving as before, I've had to make ignoreInit = TRUE, this would usually trigger the code as soon as the app starts, so I've added a clause at the start of the observeEvent to return nothing if the button hasn't been pressed, and the modal reactiveValue hasn't been made.

---

### Known issue

When on shinyapps.io, after using the reset button as we had it before, the modal buttons become unresponsive. I've tried a few things and couldn't work out why or how to fix it.

### Solution

I've changed the reset button to be a hard reset (end session and start new session) on shinyapps.io for now. This is essentially the same behaviour as the current reset button, the difference being that it takes a little longer to execute.

---

### Notes

This PR is based on a copy of development that I diverged before reverting some of these changes on the main development branch (a mess I know).

There's a bunch of code tidying that is flagged in the server file.

I couldn't figure out how to change the reactiveValue for the environment flag in shinytest, so there's no UI tests for this work. I'm not entirely happy about that, though it's something we should look into if we hook up rselenium or testServer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@
 
 ## Introduction 
 
-A shiny app to provider a user friendly way to QA data files using centralised code. This is deployed via the DfE visual studio and rsconnect subscriptions. There are three environments only accessible to DfE AD:
+A shiny app to provide a user friendly way to QA data files using centralised code. This is deployed via the DfE visual studio and rsconnect subscriptions. There are three environments only accessible to DfE AD:
 
 - Production - https://rsconnect/rsc/dfe-published-data-qa
 - Pre-production - https://rsconnect-pp/rsc/dfe-published-data-qa
--- a/README.md
+++ b/README.md
@@ -29,6 +29,10 @@ There is also an instance hosted on shinyapps.io, which is externally accessible
 
 - External - http://department-for-education.shinyapps.io/dfe-published-data-qa
 
+**Note**
+
+Sensitive data should not be screened on the shinyapps.io version, as we do not have control over this hosting server. A warning modal will appear to double check this with you before screening any files. If you're external to DfE and want to screen sensitive files, you can run a version of the app locally, see the [software requirements](#requirements) and [how to use](#how-to-use) sections below.
+
 Code used to create functions and variables called by the global, server, and UI scripts can be found in the 'R' folder.
 
 ---
@@ -36,7 +40,9 @@ Code used to create functions and variables called by the global, server, and UI
 ## Requirements
 
 
-### i. Software requirements 
+### i. Software requirements (for running locally)
+
+- Installation of R Studio 1.2.5033 or higher
 
 - Installation of R 3.6.2 or higher
 
@@ -52,10 +58,19 @@ Code used to create functions and variables called by the global, server, and UI
 
 ## How to use
 
+### Running the app locally
+
+1. Clone or download the repo. 
+
+2. Open the R project in R Studio.
+
+3. Run `renv::restore()` to install dependencies.
+
+4. Run `shiny::runApp()` to run the app locally.
 
 ### Packages
 
-Package control is handled using renv. You will likely need to run `renv::restore()` if this is your first time using the project.
+Package control is handled using renv. As in the steps above, you will need to run `renv::restore()` if this is your first time using the project.
 
 ### Tests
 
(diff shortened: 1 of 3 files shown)


server.R
```diff
@@ -91,541 +92,580 @@ server <- function(input, output, session) {
...
 
-    inputData <- values$datafile
-    inputMeta <- values$metafile
+  observeEvent(input$screenbutton | values$proceed_with_screening,
+    {
+      if (input$screenbutton == 0 && is.null(values$proceed_with_screening)) {
 
-    data <- readFile(inputData$datapath)
-    meta <- readFile(inputMeta$datapath)
+        # This ends it early
```
sarahmwong, line 142: This with the callback in the modal is so neat!! Could we also clear the inputData and inputMeta back to NULL in this step if it's not too faffy?

sarahmwong (commented):
This looks great! The environment handling is really clever and works well in shinyapps.io and locally :)

server.R
cjrace: Good thinking - I think I've managed to do this in the latest commit

cjrace:
I've updated the readme to add a description for how to run the app locally.

README.md
```diff
@@ -19,24 +19,34 @@
 
 ## Introduction 
 
-A shiny app to provider a user friendly way to QA data files using centralised code. This is deployed via the DfE visual studio and rsconnect subscriptions. There are three environments, all accessible to DfE AD:
+A shiny app to provider a user friendly way to QA data files using centralised code. This is deployed via the DfE visual studio and rsconnect subscriptions. There are three environments only accessible to DfE AD:
```
lauraselby: typo

server.R
```diff
@@ -91,541 +85,594 @@ server <- function(input, output, session) {
...
-    # Disable buttons ---------------------------------------------------------------------------------
-
-    disable("uiScreenButton")
-    disable(selector = "span[class='btn btn-default btn-file']")
+  # Confirmation modal for external use ----------------------------------------------------------
+
+  observeEvent(input$external_screenbutton, {
+    shinyalert::shinyalert(
+      title = "",
+      text = "This is the external version of the DfE data screening app. We do not control the servers that this app runs on. You should only use this to screen published or dummy data. For sensitive files, you should clone the repository and screen files locally instead.",
```
lauraselby: Could you make 'You should only use this to screen published or dummy data' stand out more? Add bold and a few line breaks maybe?

server.R
```diff
@@ -91,541 +85,594 @@ server <- function(input, output, session) {
...
+        if (data$mainFile %>% select(geographic_level) %>% distinct() %>% nrow() == 1) {
+          shinyjs::hide(selector = c(
+            "#trendy_tabs li a[data-value=geogTab]"
+          ))
+        }
+        else {
+          shinyjs::show(selector = c(
+            "#trendy_tabs li a[data-value=geogTab]"
+          ))
+        }
```
lauraselby, line 386: I think this if this was edited to take into account number of tests failing too it will fix the issue with this tab appearing even if the file has failed. Happy for this to be done via a separate PR if easier.

lauraselby: I highlighted more lines of code than this but there must be a limit! It's that area, but the if is a few lines above

README.md
cjrace: Where?

server.R
cjrace: I think there's changes coming to this via sarah's PR, so I'd ignore this for now, this was already there and just tidied with the code styling.

server.R
cjrace: Have done this in the latest commit @lauraselby

README.md
cjrace: Found and fixed in new commit

lauraselby (approved):

PR #44, opened 2021-07-14
New FY time identifiers and bug fixes

sarahmwong:
- Adding in financial year Part 1/Part 2 as allowable time identifiers 

Bug fixes from Trello:
- Stop geography tab appearing when tests fail
- Get the year-1 to be auto-selected as the comparison time for year-on-year changes tab
- Order and colour-code the geography aggregations table so mismatches appear first


cjrace (changes requested):
Could we add a test file with the financial year halves in?

Otherwise looking good to me - ran it locally and all the bugs seem fixed / improvements working 😄

cjrace (approved):

PR #47, opened 2021-07-15
development to master

cjrace:
This PR covers a handful of changes triggered by finding a couple of bugs in the QA tabs. See trello card here - https://trello.com/c/gg4fcpTj/2011-error-in-qa-app-with-spc-file.

Also picked up these two trello cards at the same time - https://trello.com/c/xign8nfP/1998-consider-adding-custom-disconnect-messages, and https://trello.com/c/iRgPlKzA/1364-caching-dependencies-for-workflows-and-ci.

This PR can be tested on our development environment (as well as locally) - https://rsconnect-pp/rsc/dev-dfe-published-data-qa/.

### 1. Handling for hyphens in variable names for the QA part of the app.

The 'What's in this file' and geography tabs were giving red error messages as the code couldn't handle variable names that contain hyphens. 

What's in this file tab error (fixed in this PR):

image

Geography tab before:

image

Geography tab after this PR:

image

The 'What's in this file' tab was fixed by removing the eval(parse()) from the filter summary code. For the geography tab I have created a custom error message to highlight what is going wrong to users.

I've added a test file and extended the QA UI tests to cover both of these scenarios.

### 2. Custom disconnect message

Whenever the app times out or disconnects, it now gives a more obvious refresh button and message to contact us if the problem persists. 

There's currently no way to differentiate between a timeout and a genuine error, so for now this should cover both bases and be more user-friendly than the standard greying out and tiny message in the bottom corner.

From this:

image

To this (with potential to further customise):

image

### 3. Tidying up

- Removed some unneccessary blank lines from UI test scripts and server.R.
- Removed some commented out code **(double check that these bits of code are okay to delete!)** from UI test scripts and server.R
- Renamed the ui test scripts to be more concise. I had the files pane quite small and they were going off the end and bugging me 🤷.
- Deleted/added/tweaked code comments, in ui.R, UI test scripts and server.R.
- Renaming shinyapps deploy workflow to match naming convention of others (so the badges match).
- Deleting country.csv and region.csv from the top level of the repo (seem to have been duplicated at some point as they also existed in the data folder).

### 4. Added the global santizer

This makes all of those red shiny error text messages into the standard one, rather than letting it show just anything. I tried giving a custom message to tell the user to contact us, though I couldn't get the standard message to be hidden. As such I've stuck with the standard message for now as I don't think it's too unreasonable to expect most users will realise that we are the app authors.

image

### 5. Adding tests around whether variables are following snake case

A test for any characters that should not be present in snake_case (anything that isn't a lowercase letter, number or underscore).

A test for variable names that start with anything other than a lowercase letter, as starting with an underscore or number can be problematic (and anythng else isn't snake_case).

Both of these have been put as advisory for now, I need to shout about it more and give analysts time before making it a hard fail.

### 6. Updating the status badges on the README

Adding status badges for CI on development branch and for the shinyapps.io deploy. The latter won't work yet, as the updated name for the workflow hasn't got through to master yet.

Updating all badges to use latest code suggestion from GitHub, I suspect they weren't working properly before.

### 7. Adding a step to the workflows to cache the renv packages

The time for the workflows had ballooned from around 3/4 minutes up to around 25/30 minutes, with the vast majority of the time being spent on restoring using renv (> 20 minutes). I did some digging and it looks like we were only caching the R setup, and not the renv packages, so I've added another step to explicitly cache the packages in renv for the project to make the workflow runs more efficient (renv restore back down to 20 seconds on some runs).


lauraselby (dismissed):
Needs a sad disconnect duck but otherwise looks fine to me

cjrace:
Have stumbled into making the santized global error message customised while looking at something else (turns out I just needed to force it with !important). Error message for 4. is now this (controlled by our CSS sheet):

image

sarahmwong (approved):
Nice fix for the sanitized message! Looks good :)

PR #48, opened 2021-07-21
Fixing geog not available tests bug

cjrace:
An analyst found a bug earlier where the na_geography, and na_geography_code tests were failing a file that should have passed.

Their scenario was that they had only LAD level data, but had country and regional columns completed. All the country cols gave the England details, and there was one Not available region, with one Not available code. I've added a cut down version of their file to the otherTests folder and added a unit test for it passes.

When running chunks of the code in the console the cause was all of the names and codes across multiple columns being merged together and then tested. This led it to incorrectly mixing the country name of England with the codes from the regional and lad columns, as shown here:

image

I've rewritten the tests to only consider the name and code columns specific to each geographic level. National/Regional/LA are ignored as we'll handle those in the harmonisation checks; RSC region only has a name column; and school level or equivalent are also ignored.

There's also a couple of other tiny commits that were already on dev that have snook in too: updating the CI to work on pushes to dev and fixing a url in the README.


lauraselby (approved):
Impressive! Looks ok to me but would be good if Sarah could review this too.

sarahmwong (approved):
Ahhh this is very slick - thank you for spotting and solving!

PR #49, opened 2021-07-26
fixing duplicate issue

cjrace:
This should separate out the names and codes by geographic_level so that scenarios where a location shares a name at multiple levels will pass.

I've not done this for code as I don't think the codes should be shared across levels?

Added a tiny cut of the file that triggered this as test data.


lauraselby (dismissed):

cjrace:
Reopening for review - had thought I'd merged this and can't think of a reason not to.

lauraselby (approved):

PR #50, opened 2021-07-29
Disconnect message and LA harmonisation check

lauraselby:
### **Custom disconnect message** 

A new function to display a custom disconnect message whenever the app disconnects: 

image

I started from the https://github.com/daattali/shinydisconnect package but edited the function to allow an image and then had to hard code the text to give more flexible formatting without a load of extra faff.

### **LA harmonisation checks**

I updated the old_la_code check to check for valid LA combinations (old_la_code, new_la_code, la_name) instead.

The LA lookup table is based on the geoportal code history database with 3 digit LA codes added from my best attempt based on recent releases in EES - these should be validated against the PDR but I can't access it at the moment and we know it's not up to date enough to show the April 2021 changes.

The error messages currently have placeholder guidance links until those sections are written and the links confirmed - I can have a go at writing these tomorrow and will update this PR if I do. We may need to consider holding back merging this to master until the guidance granny review changes are all in. 

I also deleted the new_la_code check as I didn't think it was necessary now the combination check is there, and updated the messages and check names for the existing country and region harmonisation checks.    

I updated the tests and all should be passing.


R/mainTests.r
```diff
@@ -1094,39 +1051,42 @@ not_table_tool <- function(data) {
   return(output)
 }
 
-# old_la_code -------------------------------------
-# Checking if old_la_code is always only 3 numeric digits
+# old_la_code_code -------------------------------------
```
cjrace: Not sure this is what you intended the comment to be?

R/mainTests.r
```diff
@@ -1094,39 +1051,42 @@ not_table_tool <- function(data) {
...
+# checking that la code and name combinations are valid
 
-old_la_code <- function(data) {
-  if (!"old_la_code" %in% names(data)) {
+la_combinations <- function(data) {
+  if (!"new_la_code" %in% names(data)) {
```
cjrace, line 1058: Any particular reason you switched from old code to new code for this validation?

R/mainTests.r
```diff
@@ -1094,39 +1051,42 @@ not_table_tool <- function(data) {
...
-      filter(old_la_code != ":") %>%
-      pull(old_la_code) %>%
-      .[!grepl("^[0-9]{3}$", .)]
+      filter(new_la_code != "") %>%
+      filter(new_la_code != ":") %>%
+      filter(new_la_code != "z") %>%
+      mutate(combo = paste(old_la_code, new_la_code, la_name)) %>%
+      pull(combo) %>%
+      .[!(. %in% expected_la_combinations)]
+    # .[!grepl("^[A-Z]{1}[0-9]{8}$", .)]
```
cjrace: Should this line just be deleted?

R/mainTests.r
```diff
@@ -1186,10 +1146,10 @@ region_code <- function(data) {
 # country_code -------------------------------------
```
cjrace: This heading for the outline should be updated to country_combinations

global.R
```diff
@@ -237,3 +237,131 @@ roundFiveUp <- function(x, n) {
...
+             position: fixed !important;
+             bottom: auto !important;
+             left: 50% !important;
+             padding: 0.8em 1.5em !important;
+             text-align: center !important;
+             height: auto !important;
+             opacity: 1 !important;
+             z-index: 99999 !important;
+             border-radius: 3px !important;
+             box-shadow: rgba(0, 0, 0, 0.3) 3px 3px 10px !important;
```
cjrace, line 349: Not sure there's enough !important 's in this function...

tests/testthat/test-mainTests.R
```diff
@@ -129,28 +123,28 @@ test_that("not_table_tool", {
...
+test_that("country_combinations", {
+  expect_equal(testIndividualTest(pathStart, "country_combinations"), "FAIL")
 })
 
-# country_code -------------------------------------------------------------------------------------------------------
+# country_combinations -------------------------------------------------------------------------------------------------------
 
-test_that("country_code", {
-  expect_equal(testIndividualTest(pathStart, "country_code"), "FAIL")
+test_that("country_combinations", {
```
cjrace: have you duplicated this test? not sure why it's there twice

data/las.csv
```diff
@@ -1,167 +1,167 @@
-new_la_code,la_name,date_of_operation,date_of_termination,parent_geography_code,entity_code,status
```
cjrace, line 1: Should we add a column for the type of location - you had that file that explained what a E09 and E06 were, might be useful to have in here?

cjrace (changes requested):

R/mainTests.r
lauraselby: No reason other than I like it better and think it's a more reliable code, I think any of the three could go here and have the same effect?

global.R
lauraselby: Hah I know! That bit is just as is from the initial function - with the differences between local and rsconnect making it so much harder I didn't bother testing if I could remove any to see if it still worked

tests/testthat/test-mainTests.R
lauraselby: Hmm I didn't add the second one, I just updated the check name. It does look like a duplicate though so I'll remove.

R/mainTests.r
cjrace: Thinking about it, we should probably check that all three are present? I think there's a scenario where a file could have 2 out of 3 columns and still sneak through to this point? E.g. an LAD file with only old_la_code and la_name.

Either that or we need another check adding entirely, which would make sure that if one of the la cols are present, the other 2 both are as well, like we do with region_name and region_code, that way we could leave the validation on this la_combinations function as you have it now

R/mainTests.r
lauraselby: yep, done

R/mainTests.r
lauraselby: good spot, done.

R/mainTests.r
lauraselby: It wasn't! Changed now

data/las.csv
lauraselby: I have entity_code in there which gives the E0- numbers? Would the description add anything? (I don't think they're included in the code history database)

cjrace: It's not a huge thing but I think the description would add something (fairly sure it's a trello card somewhere?). The entity_code feels redundant as you can just search the first 3 characters of the new_la_code to get the same result, whereas the description actually tells you what they mean. 

Happy for you to leave out of this PR if you don't have the file to hand.

lauraselby: I've added the descriptions though I think the entity stuff may be confusing for anyone looking at it as I don't think many people realise LAs aren't really a thing in their own right.

R/mainTests.r
lauraselby: Agree it should be a new test, I've created a trello card to cover it separately.

cjrace (dismissed):

R/mainTests.r
```diff
@@ -1158,22 +1117,21 @@ region_code <- function(data) {
       mutate(combo = paste(region_code, region_name)) %>%
       pull(combo) %>%
       .[!(. %in% expected_region_combinations)]
-    # .[!grepl("^[A-Z]{1}[0-9]{8}$", .)]
 
     if (length(invalid_values) == 0) {
       output <- list(
-        "message" = "region_code is always a 9 digit code, with one letter followed by 8 numbers, : for not available, or blank.",
+        "message" = "All region_code and region_name comninations are valid.",
```
sarahmwong: typo combinations

R/mainTests.r
```diff
@@ -1094,39 +1051,41 @@ not_table_tool <- function(data) {
...
+          "message" = paste0("The following old_la_code, new_la_code and la_name combination is invalid: '", paste0(invalid_values), "'. <br> - We do not expect any combinations outside of the <a href='https://github.com/dfe-analytical-services/dfe-published-data-qa/blob/master/data/las.csv' target='_blank'>standard geographies</a> (case sensitive), you can use our [guidance link TBC] for help creating local authority level data."),
           "result" = "FAIL"
         )
       } else {
         output <- list(
-          "message" = paste0("The following old_la_code values are invalid: '", paste0(invalid_values, collapse = "', '"), "'. <br> - old_la_code must always be a 3 digit numeric code, blank or : for not available.."),
+          "message" = paste0("The following old_la_code, new_la_code and la_name cominations are invalid: '", paste0(invalid_values, collapse = "', '"), "'. <br> - We do not expect any combinations outside of the <a href='https://github.com/dfe-analytical-services/dfe-published-data-qa/blob/master/data/las.csv' target='_blank'>standard geographies</a> (case sensitive), you can use our [guidance link TBC] for help creating local authority level data."),
```
sarahmwong: typo combinations

R/mainTests.r
```diff
@@ -1158,22 +1117,21 @@ region_code <- function(data) {
...
           "result" = "FAIL"
         )
       } else {
         output <- list(
-          "message" = paste0("The following region_code / region_name cominations are invalid: '", paste0(invalid_values, collapse = "', '"), "'. <br> - We do not expect any regions outside of those on the <a href='https://geoportal.statistics.gov.uk/datasets/regions-december-2020-en-bgc/data?geometry=-22.223%2C50.522%2C17.877%2C55.161' target='_blank'>ONS Open Geography Portal</a> (case senstive), or : for not available."),
+          "message" = paste0("The following region_code / region_name cominations are invalid: '", paste0(invalid_values, collapse = "', '"), "'. <br> - We do not expect any combinations outside of the <a href='https://github.com/dfe-analytical-services/dfe-published-data-qa/blob/master/data/regions.csv' target='_blank'>standard geographies</a> (case sensitive), you can use our [guidance link TBC] for help creating region level data."),
```
sarahmwong: typo

R/mainTests.r
```diff
@@ -1204,18 +1162,18 @@ country_code <- function(data) {
 
     if (length(invalid_values) == 0) {
       output <- list(
-        "message" = "country_code is always one of the expected ONS codes or ':' for 'Not available'.",
+        "message" = "All country_code and country_name comninations are valid.",
```
sarahmwong: typo

sarahmwong (dismissed):

cjrace (approved):

PR #51, opened 2021-07-30
Sch prov changes

cjrace:
## Overview

Our guidance has updated now school and provider level data is useable in EES. There are two scenarios where teams can now use this:

1. As a geography, where it is the only geographic_level in a file.
2. As a filter, where it is the only filter in a file.

If used as a geography and mixed with other levels, this would behave as before and still be ignored by the platform.

This then required a number of changes to our current checks...

### Adding test data and end to end unit tests

I've added a new folder in the testthat area, for sch_prov, and added extra tests to the screenFiles tests script. Here are the test files (and test cases) I used, it's worth reviewing them for anything I may have missed or have wrong:

1. School level only

This file should pass the check for ignored rows, and should not fail any of the other checks.

2. Provider level only 

This file should pass the check for ignored rows, and should not fail any of the other checks.

3. School name as only filter

This file should not fail any checks, and should always pass the following checks: ob_unit_meta, geographic_catch, overcompleted_cols.

4. School name as one of many filters

This file should fail the ob_unit_meta check (and therefore fail before getting to the main checks).

5. School name as a filter group (only one filter)

This file should not fail any checks.

6. School name as only filter and grouped

This file should not fail any checks.

7. Only 2 out of 3 school columns present

This file should fail at least one screening check.

8. A file with one filter, but school name in the data file

This file should fail at least one screening check.

9. Provider level only, with duplicate rows

This file should fail the duplicate rows check.

10. School and provider level together

This file should fail the ignored rows test.

11. School level mixed with other geographic levels

This file should pass the check for ignored rows with a note around ignored rows, and should not fail any of other the checks.

### Updating tests 

**General**

Added shhh-ing to tidy up the console when running tests or tidy_code locally.

**shinytest**

I have merged both test scripts into one UI tests script. This is because when running the tests locally the cache clears as soon as you update one shinytest script, which means it errors if you try to view any more test scripts with updates, causing you to have to rerun the tests for as many scripts as you have changes.

**testthat**

I have added extra context to the screenFiles test script to make seeing the failures and test coverage easier in the console and logs.

### Replaced not_table_tool()

I've replaced the old not_table_tool() check, with a more snazzy check called ignored_rows(), this serves the same purpose as the old test but accounts for the extra scenarios we have now and informs you have the number of rows that will be ignored by EES.

### Updated ancillary file message

Previously we had a blue box that was triggered by the not_table_tool() check and would show when a file should be uploaded as an ancillary file as all rows would be ignored by the table tool. This now triggers from the new ignored_rows() test.

### Added new note message

Based on the ignored_rows test, I've added a new blue box to show the number of rows that will be ignored by the table tool. I decided a warning was too negative, and that it should be treated as a pass with a note.

I've updated the stats on the main page accordingly in the server file.

### Reinstated geog validation for sch and prov

In a previous PR @sarahmwong had added two school level checks that were designed to look for code and name duplicates, these, and their unit tests, have now been uncommented.

The geog na and duplicate tests have also been expanded to cover school and provider level

### Updated preCheck2/geography_level_present()

This check now extends to school and provider level, checking their associated columns are present.

### Fixed la within lad test

Not sure why, but at some point when updating the tests a bug was flagging in this test so I tweaked it to fix that and make the tests pass again.

### Updated geography_matrix

In the knownVariables file we have a matrix of all geographic levels and associated columns, in line with the standards updates, I've updated the school and provider level columns.

This resulted in having two columns that were now completely NA, so I've removed those, and adjusted checks throughout the screener to match.

I've also updated references to geographic levels or columns throughout the code to be direct to this matrix, which should make maintenance and future updates easier. Not all references were changed as some required more effort to change and I decided they weren't worth it (yet!).

### Removed redundant tests

In the last day @lauraselby has added an la combinations check which renders the old_la_code and new_la_code validation redundant, so I've removed those checks, and associated tests and test data.

### Updated preCheck2/ob_unit_meta

This check tests for observational units across the metadata and flags an error, I've updated this to allow school_name and provider_name where they are they only filter.

I've also expanded this check to catch filter groups as well, which appears to have been an omission previously.

### Updated geographic_catch

This check uses the regex in the knownVariables file to catch likely geographic columns in the metadata, like the ob_unit_meta check above, I've adapted this to cover the new scenarios, and also expanded to include the filter groups.

### Updated overcompleted_cols

My favourite check... I've updated this to account for the new scenarios, to all school_name or provider_name to be completed for other geographic levels where they are a filter (or filter group).

### Updated ob_unit_total

This check catches total values in observational units, so I've had to adapt it to catch them in the sch and prov columns, but not where they are filters.

I may not have covered this one fully, I'll open up the PR but is definitely worth double checking.

### Updated duplicate_rows

Before, this test ignored the school and provider rows when looking for duplicates to match EES. Now it will ignore them only if they are not the only geographic level.


R/knownVariables.r
```diff
@@ -31,54 +31,49 @@ four_digit_identifiers <- c("Calendar year", "Reporting year", weeks, months)
...
+  "Local enterprise partnership", "local_enterprise_partnership_code", "local_enterprise_partnership_name", NA,
+  "English devolved area", "english_devolved_area_code", "english_devolved_area_name", NA,
+  "Opportunity area", "opportunity_area_code", "opportunity_area_name", NA,
+  "Ward", "ward_code", "ward_name", NA,
+  "MAT", "trust_id", "trust_name", NA,
+  "Sponsor", "sponsor_id", "sponsor_name", NA,
+  "School", "school_urn", "school_name", "school_laestab",
+  "Provider", "provider_ukprn", "provider_name", NA,
+  "Institution", "institution_id", "institution_name", NA,
+  "Planning area", "planning_area_name", "planning_area_code", NA
```
sarahmwong: Very minor/feel free to ignore but can we switch positions of planning_area_name and planning_area_code just so they match up with the rest of the matrix cols in case we need to pull just names out in future?

tests/testthat/test-screenFiles.R
```diff
@@ -1,4 +1,4 @@
-context("screenFiles")
+context("No filters")
```
sarahmwong, line 1: These are so neat!

R/mainTests.r
```diff
@@ -135,22 +135,32 @@ duplicate_rows <- function(data, meta) {
...
+      filter(geographic_level != geography_matrix[15, 1]) %>%
+      filter(geographic_level != geography_matrix[16, 1]) %>%
+      select(present_obUnits_filters) %>%
+      get_dupes())
+  }
 
   if (nrow(dupes) > 0) {
     output <- list(
-      "message" = paste("There are", cs_num(nrow(dupes)), "duplicate rows in the data file. <br> - Note that school, provider, institution, and planning area level rows are not included in this test."),
+      "message" = paste("There are", cs_num(nrow(dupes)), "duplicate rows in the data file. <br> - Note that,", paste0(geography_matrix[13, 1], ", ", geography_matrix[14, 1], ", ", geography_matrix[15, 1], " and ", geography_matrix[16, 1]), "level rows are not included in this test."),
```
sarahmwong: Could add another message for when you screen a school/provider-only level file to say it's ignoring only institution and planning area - at the moment says it ignores the provider/school level rows but flags as a fail for that provider test file with only provider-level data in it
image

R/mainTests.r
```diff
@@ -1042,12 +1100,44 @@ not_table_tool <- function(data) {
...
 
+      if (nrow(levels_present) == 1 && data$geographic_level[1] %in% geography_matrix[13:14, 1]) {
+        output <- list(
+          "message" = "No rows in the file will be ignored by the EES table tool.",
+          "result" = "PASS"
+        )
+      } else {
+        if (geography_matrix[13, 1] %in% levels_present$geographic_level && geography_matrix[14, 1] %in% levels_present$geographic_level) {
+          output <- list(
+            "message" = paste(geography_matrix[13, 1], "and", geography_matrix[13, 1], "data has been mixed - please contact the Statistics Development Team."),
```
lauraselby: should these be different?

sarahmwong: One needs pointing to "Provider" at geography_matrix[14,1] as currently repeats "School"
image

cjrace: Yes, good spot!

R/mainTests.r
```diff
@@ -1232,13 +1306,13 @@ school_laestab_duplicates <- function(data) {
...
-        "message" = "Every school name has one laestab, and vice versa.",
+        "message" = paste("Every", geography_matrix[13, 3], "value has one", geography_matrix[13, 4], "value, and vice versa."),
         "result" = "PASS"
       )
     } else {
       if (length(multi_count_code) == 1 & length(multi_count_name) == 0) {
         output <- list(
-          "message" = paste0("The following school_name has multiple laestabs: ", paste0(multi_count_code), ". 
+          "message" = paste0("The following ", geography_matrix[13, 3], "value has multuple", geography_matrix[13, 4], " values: ", paste0(multi_count_code), ". 
                              <br> - Every school name should have only one laestab and vice versa."),
```
sarahmwong: Potential issue with this check but not sure if I'm just overthinking! - e.g. we can have two legitimate "St Paul's" schools in different parts of the country with different LAESTABs - way around it is taking into account LA? But then that forces analysts to include that level info in all school files. Might be easier to just remove the part of the test that checks for name sharing a code for now? Codes sharing a name should still be highlighted though

R/mainTests.r
```diff
@@ -1292,27 +1366,13 @@ school_laestab_duplicates <- function(data) {
 
 # school_urn_duplicates  ----------------------------------------
 # check that there is a 1:1 relationship between school urns and names
+
```
sarahmwong: same as above for legitimate duplicate code sharing

sarahmwong (commented):
Looking fab - really neat bits with the context for screening files! Added a couple comments around the school tests as think this might accidentally highlight some legit cases that should be allowed in EES (know I wrote these before! So happy to take a look if you like :) )

tests/testthat/test-screenFiles.R
cjrace: I was so happy when I realised we could do this!

R/knownVariables.r
cjrace: Good spot! I'll fix in the next commit

R/mainTests.r
cjrace: Agree - will fix in the next commit

R/mainTests.r
cjrace: Same as Laura's comment - think I've just pushed a fix

R/mainTests.r
cjrace: As we're using LAEstab we could extract the LA code from the first 3 digits of it to identify the LA. I guess we could remove the duplicate name part of these checks? There shouldn't be shared codes I don't think? Can discuss on stand-up with @lauraselby

R/mainTests.r
cjrace: Don't think there should be any duplicate URN's, think they will always be an error?

R/mainTests.r
cjrace: Resolving this as now removing these checks entirely (the platform should be fine with duplicates like these)

R/mainTests.r
cjrace: Resolving this as now removing these checks entirely (the platform should be fine with duplicates like these)

cjrace:
I've removed the school and provider duplicate name / code tests as I feel they don't make as much sense anymore. EES should in theory be able to handle that data (or be amended to do so), and there's a number of legit cases we'd rule out. Better to wait until we harmonise that data and define the lookups we want teams to use and then check against that.

cjrace:
From trello I've added a check for where if one (or two) of the la columns is either present or completed but at least one of the three isn't, then it will flag an advisory explaining this. This is to encourage people to use all three columns rather than just one or two of them.

I've also fixed the custom validation errors for the inputs in the qa tabs, they now use the shinyFeedback package like the upload validation errors, or sweetalerts. Our custom blanket error message is preserved and doesn't overwrite these need messages (unlike the previous `validate(need())` ones we were using). They're not fully reactive, so the errors don't clear until you click the submit button again, though even with this, these changes significantly improve the user experience for now. 

I had a check and shinytest doesn't seem to cover these unless we use screenshots, so I've left the tests as is for now and put a card on trello for us to review the tests at some point and look at adding something fancier like selenium to see if we can get that to recognise these without screenshots. 

Screenshots of the validation I've added:

image
image
image
image
image
image

server.R
```diff
@@ -707,12 +707,26 @@ server <- function(input, output, session) {
 
           # create a list of tables - with one for each indicator summary
           theList <- eventReactive(input$submit, {
-            validate(
-              need(input$ind_parameter != "", "Please select at least one indicator"),
-              need(input$geog_parameter != "", "Please select at least one geographic level")
-            )
+            if (is.null(input$ind_parameter)) {
+              shinyFeedback::showFeedbackDanger(
```
sarahmwong, line 711: Ah this is such a neat workaround! Looks really good on the app

tests/shinytest/UI_tests-expected/022.json
```diff
@@ -1,9 +1,9 @@
 {
   "output": {
     "geog_agg2": [
-      "You have at least one hyphen in your variable names, you need to remove all hyphens from variable names to use this part of the app.",
+      "An error has occurred. Check your logs or contact the app author for clarification.",
```
sarahmwong, line 4: Correct error message shows when I run the shinytest file through the app - is it just showing this because we've overwritten this default message?
image

R/mainTests.r
```diff
@@ -898,6 +899,120 @@ region_col_completed <- function(data) {
...
+        } else {
+          if (!("new_la_code" %in% names(data)) && !("old_la_code" %in% names(data))) {
+            output <- list(
+              "message" = "Where la_name is included in the data file, new_la_code and old_la_code should also be included.",
+              "result" = "FAIL"
+            )
+          } else {
+            if (!("new_la_code" %in% names(data))) {
+              output <- list(
+                "message" = "Where la_name and old_la_code are included in the data file, new_la_code should also be included.",
```
sarahmwong, line 900: Our guidance page advises to leave cells blank if using legacy LAs without a 9-digit code - will we need to change to suggest they use ":" instead? (or now that we have the LA lookup file, could remove that bit of guidance entirely?)

tests/shinytest/UI_tests-expected/022.json
cjrace: Yeah, shiny test isn't really doing much for these new error messages now which is frustrating. I had a look through all the outputs they take but couldn't find anything useful so just left it as this for now, I've added something to trello to look into changing our testing around the QA tabs, maybe looking at rselenium or using specific screenshots.

R/mainTests.r
cjrace: Ah yes of course!

In the latest commit I've stripped out the 'completed' part of this test, so now it's just a check for all columns being there if any are. I've done the same for the  region_completed test too.

cjrace: Sorry - to be clear, the guidance is still good to stay as it is, I've just changed this check and the region_completed check to match it.

I've also done a fix to the la_combinations test that the new test data flagged (it would fall over if some, but not all, la columns were present).

sarahmwong (approved):
Looks great to me :)

PR #53, opened 2021-08-05
Sch + provider updates, plus others

cjrace:
Including:

1. All of the school and provider changes as per #51 
2. Refactoring of UI tests to single script, and misc unit tests for better labelling
3. Cleaning up/shushing verbose console messages for tidying code and running tests locally
4. Fixing our custom validation error messages in the QA tabs
5. Amending `la_combinations()` so it doesn't break where only some LA cols are present
6. Amending `region_completed()` and `la_completed()` to `x_present()`, removing the redundant completed checks now we have a lookup to cross reference against


cjrace:
Have added in a temporary check for name duplicates in school and provider data to cover EES from this data until we have a longer term fix in.

sarahmwong (commented):
Looks good! Happy to approve - just double checking it's only relying on URN, so duplicate LAESTABs don't matter?

cjrace:
@sarahmwong yeah that's intentional, EES will ignore LAEstab

sarahmwong (approved):

lauraselby (approved):

PR #54, opened 2021-08-23
Adding outside of england to screener

cjrace:
Added it into the acceptable regions csv, and also some test data along with a unit test for it.

### DIFF
--- a/data/regions.csv
+++ b/data/regions.csv
@@ -10,3 +10,4 @@ E12000008,South East
 E12000009,South West
 E13000001,Inner London
 E13000002,Outer London
+z,Outside of England and unknown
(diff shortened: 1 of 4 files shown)


data/regions.csv
```diff
@@ -10,3 +10,4 @@ E12000008,South East
 E12000009,South West
 E13000001,Inner London
 E13000002,Outer London
+z,Outside of England and unknown
```
lauraselby, line 13: Are we really expecting the label to be 'Outside of England and unknown'? Should they not be split and just both share the z code?

cjrace: Yes, that is the specific aggregation they have always used with the name of 'Other'. Even if they did split it up (which we discussed) they should still supply the combination of them all as users expect to see it.

sarahmwong (dismissed):

sarahmwong (commented):
Sorry know I just approved a sec ago! Just ran through on dev with some other names for "z" and it passes all checks is that an issue for EES if things share a code?
image

cjrace:
@sarahmwong very good spot! Though I suspect this may already have been an issue before this PR, can you test that file on master too? We control Not available quite closely, but haven't yet put any similar limits onto Not applicable (off the top of my head). 

If it's as I suspect then I'll stick something on trello as we should probably treat them both the same, and then add additional things in (like the one in this PR) by exception.

sarahmwong:
> @sarahmwong very good spot! Though I suspect this may already have been an issue before this PR, can you test that file on master too? We control Not available quite closely, but haven't yet put any similar limits onto Not applicable (off the top of my head).
> 
> If it's as I suspect then I'll stick something on trello as we should probably treat them both the same, and then add additional things in (like the one in this PR) by exception.

Ah yes same thing happening on master - good plan! Will approve this PR for now as it will match the guidance PR :)

sarahmwong (approved):

lauraselby (approved):

PR #57, opened 2021-10-29
Bug fixes

sarahmwong:
- Reorder shinyJS hide/show calls so missing geographic level doesn't crash the app
- Get screener to read in date as a character, so any tests on date columns don't crash the app
- Add UI tests for these files to check the app doesn't crash with either case
- Hide the YoY comparisons tab if the time identifier is not a year

Last bullet is dependent on shinyJS - can't run UI tests on when tabs appear/disappear yet but want to get this in with selenium tests in future


server.R
```diff
@@ -382,15 +415,31 @@ server <- function(input, output, session) {
           shinyjs::hide(selector = c(
             "#trendy_tabs li a[data-value=geogTab]"
           ))
-        } else {
-          shinyjs::hide(selector = c(
+        }
+        # hide yoy changes if data is not split by year
+        else if (failed_tests == 0 & data$mainFile %>%
+          select(time_identifier) %>%
+          filter(!time_identifier %in% four_digit_identifiers[1:2]) %>%
```
cjrace: This seems to exclude 6 digit years like academic - should we add those back in?

sarahmwong: great spot thank you! Pushed tidy changes up now :)

cjrace (approved):

PR #58, opened 2021-11-01
Bug fix: filter levels

sarahmwong:
Fixing bug where filter levels were not displaying. Caused by handling error where there was a mix of filters with and without filter groups. Added UI tests in to catch this fix, and also added pages to the outputs (previously cut off at 10 rows).


cjrace (approved):
Looks good to me!

PR #61, opened 2021-11-09
amend code/name check to ignore blanks

sarahmwong:
Fix bug where lower level geographies counted as duplicates


sarahmwong:
Closing pull request as issue appeared to be corrupted file rather than anything else!

PR #63, opened 2022-01-18
Filter interactions

sarahmwong:
Adding in the filter interaction checks to screener. Now gives option to download a file of all the interactions that could possibly exist, but don't exist in your file.

NB: Still have the issue of large numbers of filters crashing at this stage. They still can pass the screener, but the app will stall if they go to download their file as R simply can't handle the number of possible combinations in some cases (e.g. the KS4 release with 14 different filter columns, running in R to find the potential column combos crashes regardless of it being in the app).

Gone with this solution that works for most files and gives teams enough detail as a download if they click the button, without overloading the app with generating this every time. Other option is to just give a simple "X out of X (x%) of potential combinations are in your file", but then misses the useful detail of exactly which combinations don't exist.


server.R
```diff
@@ -561,43 +561,60 @@ server <- function(input, output, session) {
...
-          #     filter(is.na(flag)) %>%
-          #     select(-flag)
-          #
-          # })
-          #
+
+
+          output$downloadFilterPerms <- downloadHandler(
+            filename = function() {
+              paste("missing_filter_combinations_", basename(inputData$name), ".csv", sep = "")
```
cjrace: I think we can drop the ".csv" here as it makes it '.csv.csv'

cjrace (commented):
Wonder if actually we should include that % as well as a line above the download button? 

Just tested it with the passes everything file and when you download you just get this, which makes sense, but could be confusing.

image

server.R
sarahmwong: Yeah that % line would make cases like this a lot clearer! Will look at adding that in now - thanks :)

sarahmwong:
test failing here because when you upload a file, then reset and upload another, it auto populates and generates the download file. When you press the button to generate filter group mixes too, it generates a line for the first file you uploaded and not the most recent one (e.g. this for filter_mix_groups.csv pops up, which is the message for passes_everything.csv):

image

I've been playing around with placement of the observeEvent(input$filter_combos, {}) and also the resetting of buttons/outputs when you click the reset button, but still having no luck. Going to convert this PR to draft for now but I think the solution is somewhere around the reactivity!

Linked to this issue on Trello here I think > https://trello.com/c/HVwHG8MX/2008-explore-data-and-yoy-changes-seem-to-be-trying-to-query-immediately-when-using-files-after-reset but more of an issue because the process for generating the file is memory-intensive, so if you upload a file with loads of filters, you won't want it to auto generate as this will crash the app

PR #65, opened 2022-03-03
GSS symbols guidance update

rmbielby:
GSS have updated their guidance on symbols, changing the Not Available symbol from **:** to **x** and the Rounds to zero symbol/code from **~** to **low**. The screener has been updated to allow these in data entries and _require_ **x** for geographical codes where not available. In addition, the screener also now looks for any obsolete symbols, not just the obsolete suppression symbol (which happened to be **x**, so has been reincarnated as Not Available).

Tests and test files in testthat have been updated accordingly.


R/knownVariables.r
```diff
@@ -30,6 +30,15 @@ four_digit_identifiers <- c("Calendar year", "Reporting year", weeks, months)
 
 six_digit_identifiers <- c("Autumn and spring term", terms, "Financial year", "Academic year", "Tax year", financial_quarters, financial_halves)
 
+# Latest GSS/ONS data symbols
+gssNAvcode <- "x"
+gssNApcode <- "z"
+gssSupcode <- "c"
+gssRndcode <- "low"
+Symbol <- (c(gssNApcode, gssNAvcode, gssSupcode, gssRndcode))
```
cjrace: `Symbol` should be more informative so it makes more sense in the other code scripts, e.g. `gss_symbols`

cjrace: Also - is there a reason for having the extra brackets around the c() like (c()) ?

R/mainTests.r
```diff
@@ -1200,8 +1200,7 @@ eda_combinations <- function(data) {
       unique() %>%
       filter(!is.na(.)) %>%
       filter(english_devolved_area_code != "") %>%
-      filter(english_devolved_area_code != ":") %>%
-      filter(english_devolved_area_code != "z") %>%
+      filter(!(english_devolved_area_code %in% Symbol)) %>%
```
cjrace: I don't think we want to filter out all symbols. Something like 'low' we'd never expect in the geography columns. Could either limit it to fewer symbols, or drop this and add Not available and it's symbol as a name/code combo in the lookup file so that it's caught there?

cjrace: This repeats across the other geography checks which do the same

cjrace (changes requested):
Looks good, just a few small comments added on the code.

Also the tidy code checks are failing so looks like you need to run tidy_code() before your next push to make sure it's all formatted cleanly.

R/mainTests.r
rmbielby: How about the Not applicable symbol? The "z" was there previously, but does it need to be? Should it only be "x" here?

cjrace: The more I think about it the more I think we should add the Not available / x combination to the lookup csv's and then stop filtering out codes here. These tests were written before we added the lookups and filtering out by code opens it up to using an incorrect name with a code. This is linked to this card on our trello board - https://trello.com/c/hxjJfrj6/2270-z-with-any-name-passes-the-screener-add-more-checks-to-lock-it-down-like

tests/testthat/mainTests/roundedzero_symbols.csv
```diff
@@ -0,0 +1,376 @@
...
+201718,Academic year,Regional,E92000001,England,E12000003,Yorkshire and the Humber,,,,State-funded primary and secondary,2088,693768,4.45385,2.93793,1.51592,82336,11.86794
+201718,Academic year,Regional,E92000001,England,E12000003,Yorkshire and the Humber,,,,State-funded primary,1771,394967,3.85078,2.64836,1.20243,40339,10.21326
+201718,Academic year,Regional,E92000001,England,E12000003,Yorkshire and the Humber,,,,State-funded secondary,317,298801,5.26744,3.32858,1.93886,41997,14.05517
+201718,Academic year,Regional,E92000001,England,E12000004,East Midlands,,,,State-funded primary and secondary,1921,592559,4.15794,3.00161,1.15633,62888,10.61295
+201718,Academic year,Regional,E92000001,England,E12000004,East Midlands,,,,State-funded primary,1633,338065,3.63591,2.65821,0.9777,~,9.17723
+201718,Academic year,Regional,E92000001,England,E12000004,East Midlands,,,,State-funded secondary,288,254494,4.85183,3.45806,1.39377,31863,12.52014
+201718,Academic year,Regional,E92000001,England,E12000005,West Midlands,,,,State-funded primary and secondary,2185,767407,4.20781,2.97631,1.2315,82604,10.76404
```
cjrace, line 17: Should drop any rows we don't need so that this file is like 10-15 rows, saves the repo bloating too much as more test data is added

cjrace:
Spotted by accident this morning - we should update our lookup files in the /data folder to use 'x' rather than ':' for any codes that are not available

cjrace (approved):
Happy to approve now - thanks for the changes! Also like the comment in the knownVariables file, good thinking to highlight where else that symbol is hardcoded!

PR #70, opened 2022-06-06
Added "Outside of England and unkown" to permissiable LAs file.

rmbielby:
Quick update to add z/Outside of England and unkown to the LAs list.

### DIFF
--- a/R/fileValidation.r
+++ b/R/fileValidation.r
@@ -2,7 +2,8 @@
 # File validation functions
 
 fileValidation <- function(datafilename, metafilename, dataseparator, metaseparator, datafile, metafile) {
-  as_tibble(t(rbind(cbind(
+  print(datafilename)
+  x <- as_tibble(t(rbind(cbind(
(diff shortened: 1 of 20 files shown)


cjrace (changes requested):
Addition looks fine, though probably worth sticking a test file that includes it in as a part of the additional geography unit tests so we're happy that the file passes.

Not sure why you've added the print() statements, were they just for debugging? Think they should be removed.

Tests and tidy code are failing / need resolving.

Also worth merging my dependency updates branch into this, checking it still works with the new packages, and then that other PR will close automatically when this one goes in too.

rmbielby:
Added test case into the existing test case file: 
[adding_z_locationCode.csv
](tests/testthat/otherData/adding_z_locationCode.csv)

Also added z's into hierarchy csv file and into country.csv as it seemed to make sense that they should be in there too.

cjrace (approved):

PR #76, opened 2022-12-21
Ethnicity harmonisation

rmbielby:
## Overview

Added in ethnicity checks comparing against the GSS standards and our ethnicity guidance.

I've added the following four tests:
- ethnicity_headers(metafile)
- ethnicity_values(datafile)
- ethnicity_characteristics_group(datafile)
- ethnicity_characteristics_values(datafile)

The first two check for any specific ethnicity filters and then check the field labels and filter values against our guidelines. It'll return an fail if the field labels aren't as outlined in the guidance and a warning if the values don't match.

The second two do the same but for general characteristics filters. A fail will be raised if any characteristic_group values containing "ethnic" don't match our guidelines and then any values in characteristic that are contained in an ethnic characteristic group that don't conform to our guidance will also be flagged with a warning.


R/mainTests.r
```diff
@@ -1417,12 +1421,12 @@ region_combinations <- function(data) {
         filter(geographic_level == geography_matrix[2, 1]) %>%
         select(geography_matrix[2, 2], geography_matrix[2, 3]) %>%
         unique() %>%
-        filter(region_code != gssNAvcode && !is.na(region_code)),
+        filter(region_code != gssNAvcode | is.na(region_code)),
```
chfoster, line 1418: just double check - was this an intended change (they look like they do different things but idk)

R/mainTests.r
```diff
@@ -2539,3 +2543,194 @@ indicator_dp_completed <- function(meta) {
...
+        "The ethnicity filter ", value_type, " '",
+        paste(ethnicity_nonstandard, collapse = "', '"),
+        "' does not conform to the GSS standards. Please cross check against the <a href='https://rsconnect/rsc/stats-production-guidance/ud.html#Ethnicity'>published standards</a>."
+      ),
+      "result" = "ADVISORY"
+    )
+  } else {
+    output <- list(
+      "message" = paste0(
+        "The following ethnicity filter ", value_type, "s do not conform to the GSS standards: <br> - '",
```
chfoster, line 2631: I think this should be filter**s** before the value_type (typo, s is currently after?)

tests/testthat/mainTests/ethnicity_characteristics_filter_wrong_group_and_values.csv
```diff
@@ -0,0 +1,958 @@
+"time_period","time_identifier","geographic_level","country_code","country_name","region_code","region_name","characteristic_group","characteristic","full_time_equivalent","headcount"
```
chfoster, line 1: This file is missing a .meta file?

tests/testthat/mainTests/ethnicity_characteristics_filter_wrong_values.csv
```diff
@@ -0,0 +1,958 @@
+"time_period","time_identifier","geographic_level","country_code","country_name","region_code","region_name","characteristic_group","characteristic","full_time_equivalent","headcount"
```
chfoster, line 1: This file is missing a .meta file?

chfoster (changes requested):

R/mainTests.r
rmbielby: The variable value_type can be either "combination" or "value", so the above should get rendered as "...filter values do not..." or "...filter combinations do not..."

chfoster: Fair!!

tests/testthat/mainTests/ethnicity_characteristics_filter_wrong_group_and_values.csv
rmbielby: Good spot, added in now and expanded test set.

tests/testthat/mainTests/ethnicity_characteristics_filter_wrong_values.csv
rmbielby: As above, added in and expanded unit tests to cover.

R/mainTests.r
rmbielby: Checked back and this was down to a change in behaviour of the code that reads in the data. It used to read blank entries as empty strings, but now reads them in as NAs. This bit of code was originally searching for the empty strings to identify blanks and this change allows it to still find the blanks by searching for NAs. There shouldn't be any valid reason for an NA in the region name or code fields if the geographic level is set to regional, so think this is safe. Good to have it described here now, so thanks for highlighting.

PR #84, opened 2023-03-15
Characteristics filter screening

rmbielby:
Have added a test to look for the characteristic and/or characteristic_group filters in the meta data file. It flags an advisory warning if they're present and a pass if not.

Have tested it with a sample file, but not created an automated test for it. The additional output is present in the shiny-test snapshots though.


cjrace:
To fix R dependency issue in automated pipelines and look for a test file to use for it

R/mainTests.r
```diff
@@ -130,6 +131,26 @@ variable_start_letter <- function(data) {
...
 }
 
+# variable_snake_case -------------------------------------
+# Checking datafile for presence of characteristic_group or characteristic filter fields
+
+variable_characteristic <- function(meta) {
+
+  if ('characteristic' %in% meta$colname | 'characteristic_group' %in% meta$colname | 'characteristic_group' %in% meta$filter_grouping_column) {
+    output <- list(
+      "message" = "The fields characteristic and/or characteristic_group have been included in the data. These are not recommended for use with the EES Table Tool and instead filters should be broken out into individual fields.",
```
cjrace: Add link to guidance

rmbielby: Done.

rmbielby:
> To fix R dependency issue in automated pipelines and look for a test file to use for it

The R dependency issue we saw earlier seems to have gone away now, just a glitch on the last run.

Have added test in testthat for the new check.

cjrace:
Should point this to dev first too @rmbielby (as well as fixing tidy code) to save needing to merge main back into dev after

PR #94, opened 2023-08-08
renv and package update

rmbielby:
Deploy failing on png 0.1-7, so taking the chance to do a full package update, including boosting to renv 1.0


rmbielby:
Made an update in Find PhantonJS path part of the Shiny Test script that might be about to break the rest of the dashboard shiny test scripts as well. Associated with deprecation of previous method for storing variables during GitHub Actions:
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

Was running this at the same time as shiny template as well though, which also has the deprecated line, but didn't fail to run.

chfoster:
I've just tried running this locally and when I click 'screen files; it crashes with this console error? 
image

rmbielby:
> I've just tried running this locally and when I click 'screen files; it crashes with this console error? image

I suspect that's more to do with an issue with renv on your copy than the repo itself. Can you try restoring or re-initialising renv in your local version and see if that makes a difference?

chfoster:
> > I've just tried running this locally and when I click 'screen files; it crashes with this console error? image
> 
> I suspect that's more to do with an issue with renv on your copy than the repo itself. Can you try restoring or re-initialising renv in your local version and see if that makes a difference?

lmao yeah that's working now!

chfoster (approved):

PR #96, opened 2023-10-23
Terminate Cumbria

rmbielby:
Cumbria should have been listed as terminated in the last geography update as it was replaced by Cumberland + Westmoreland and Fenland. Making this update now.

Also updated renv whilst I was in the repo.

And another thing: bringing combined LADs within GSS guidance:
https://geoportal.statistics.gov.uk/datasets/ons::guide-to-presenting-statistics-for-administrative-geographies-may-2023/about

One more thing... shiny tests were being a little awkward. Turned out to be delays needed after button clicks and struggling with the spinner. Did a little clean up in the dashboard code and the tests to get the UI tests up and running again.

---
I branched off master in the first instance, which looks like it'd gotten ahead of the development branch, so this effectively merges the latest master into development as well.


data/lads.csv
```diff
@@ -344,5 +344,5 @@ E07000100,St Albans,,2011-12-01,terminated
 E07000101,Stevenage,,2011-12-01,terminated
 E07000104,Welwyn Hatfield,,2011-12-01,terminated
 E08000020,Gateshead,,2011-12-01,terminated
-z,Fenland and East Cambridgeshire combined,,,
+E07000010 / E07000009,Fenland / East Cambridgeshire combined,,,
```
cjrace: Should this still have 'combined' in? Does the GSS guidance say either way?

renv.lock
```diff
@@ -1,10 +1,10 @@
 {
   "R": {
-    "Version": "4.2.2",
+    "Version": "4.3.1",
     "Repositories": [
       {
         "Name": "CRAN",
-        "URL": "https://cran.rstudio.com"
+        "URL": "https://packagemanager.posit.co/cran/latest"
```
cjrace, line 7: Interesting - did it shift this itself, or is this something we should be doing with other repos too?

renv.lock
```diff
@@ -331,20 +333,23 @@
     },
     "config": {
       "Package": "config",
-      "Version": "0.3.1",
+      "Version": "0.3.2",
       "Source": "Repository",
-      "Repository": "CRAN",
+      "Repository": "RSPM",
```
cjrace, line 338: What does RSPM stand for?

cjrace: R Studio Package Manager?

cjrace (changes requested):
A few comments / questions out of interest. 

We should try to get the tests working before we merge this to master, though happy for that to happen on development rather than here.

renv.lock
rmbielby: It did that one itself

data/lads.csv
rmbielby: Good spot, meant to leave that off. Have removed it now. Here's the relevant guidance that covers this: 

> Any statistical publication using combined geographical areas should define these areas clearly, in terms of their names and geography codes, for example Cornwall/Isles of Scilly, E06000052/E06000053.

server.R
```diff
@@ -798,7 +787,7 @@ server <- function(input, output, session) {
               t_list <- purrr::imap(theList(), ~ {
                 tagList(
                   h4(.y),
-                  DTOutput(outputId = paste0("t_", .y), width = "100%") %>% withSpinner()
+                  DTOutput(outputId = paste0("t_", .y), width = "100%")
```
cjrace, line 790: Why is the spinner removed?

ui.R
```diff
@@ -342,11 +342,21 @@ fluidPage(
                     fluidRow(
                       column(
                         4,
-                        uiOutput("indicatorChoice") %>% withSpinner(),
+                        selectInput(
+                          inputId = "ind_parameter",
+                          label = "Choose indicator(s):",
+                          choices = "",
+                          multiple = TRUE
+                        )
```
cjrace, line 350: Was it intentional to remove the spinner?

ui.R
```diff
@@ -342,11 +342,21 @@ fluidPage(
...
                       ),
                       column(
                         4,
-                        uiOutput("geogChoice") %>% withSpinner()
+                        selectInput(
```
cjrace, line 354: Was it intentional to remove the spinner?

cjrace:
Not made the link with the right bit of code in my head yet, though looks like something around what boxes appear when has changed. So the same file on this branch locally (left) now shows an empty box that doesn't appear if you do it on main (right). 

Something I'd hope UI tests would catch, potentially either not set up to right now, or they are and it's just lost in the noise

image

cjrace:
Have tested it and seems fine without the spinners. Couple of odd bits that aren't the best UX wise, but as it's all in the optional extras after it passing that we care less about, and tbh, want refactoring / improving anyway, then I think removing the spinners is fine. Just the issue with the information box appearing to diagnose and fix, then it's good to go @rmbielby (and update the test snapshots after the URL changes in your recent commit!)

cjrace:
> Have tested it and seems fine without the spinners. Couple of odd bits that aren't the best UX wise, but as it's all in the optional extras after it passing that we care less about, and tbh, want refactoring / improving anyway, then I think removing the spinners is fine. Just the issue with the information box appearing to diagnose and fix, then it's good to go @rmbielby (and update the test snapshots after the URL changes in your recent commit!)

Tried again and couldn't recreate the information box issue, not sure what happened but writing it off for now unless it appears on one of the environments

cjrace (approved):

rmbielby:
> Not made the link with the right bit of code in my head yet, though looks like something around what boxes appear when has changed. So the same file on this branch locally (left) now shows an empty box that doesn't appear if you do it on main (right).
> 
> Something I'd hope UI tests would catch, potentially either not set up to right now, or they are and it's just lost in the noise
> 
> image

Just a note that we couldn't reproduce this issue again, so writing off as a one off glitch for now.

PR #98, opened 2023-11-01
Adding Local skills improvement plan area as a geographic level

rmbielby:
Local skills improvement plan areas (LSIPs) are required by the Further Education publication team as a searchable geographic level. Explore Education Statistics now accepts LSIPs as a geographic level and this PR updates the EES data screener to match what is allowed.

The behaviour around LSIPs is:
- Local skills improvement plan area is an allowable geographic level
- Any local skills improvement plan area data rows must contain valid entries under lsip_code and lsip_name in accordance with the supplied codes and names file (lsips.csv)
- lsip_code and lsip_name are also allowed to be present in LAD geographic level rows

To accomplish the above, LSIPS are checked as part of:
- lsip_combinations()
- overcompleted_cols()

To check continued working of the screener with LSIPs, I've amended/added the following tests of the screener itself:
- Added a specific test (with failure) of the lsip_combinations() check.
- The passes_everything test checks that correct lsip entries pass all tests


data/lsips.csv
```diff
@@ -0,0 +1,39 @@
+lsip_code,lsip_name
```
cjrace, line 1: Would it be worth adding a section to the readme to list out data sources, could start it just with LSIPs for where we've got this from, and then do in a separate PR to add in the equivalent source information for our other lookups?

R/mainTests.r
```diff
@@ -168,42 +169,42 @@ duplicate_rows <- function(data, meta) {
...
       filter(geographic_level != geography_matrix[15, 1]) %>%
       filter(geographic_level != geography_matrix[16, 1]) %>%
+      filter(geographic_level != geography_matrix[17, 1]) %>%
       select(all_of(present_obUnits_filters)) %>%
       get_dupes())
 
     if (nrow(dupes) > 0) {
       output <- list(
-        "message" = paste("There are", cs_num(nrow(dupes)), "duplicate rows in the data file. <br> - Note that", paste0(geography_matrix[13, 1], ", ", geography_matrix[14, 1], ", ", geography_matrix[15, 1], " and ", geography_matrix[16, 1]), "level rows were not included in this test."),
+        "message" = paste("There are", cs_num(nrow(dupes)), "duplicate rows in the data file. <br> - Note that", paste0(geography_matrix[14, 1], ", ", geography_matrix[15, 1], ", ", geography_matrix[15, 1], " and ", geography_matrix[16, 1]), "level rows were not included in this test."),
```
cjrace: Think you missed some ID changes on this line? Has 15 twice in a row, should it be 14/15/16/17?

tests/testthat/test-screenFiles.R
```diff
@@ -17,7 +17,7 @@ context("Passes everything")
 test_that("PassEverything", {
   screeningOutput <- testOther("../../tests/shinytest/test-data/passes_everything.csv")
 
-  expect_equal(screeningOutput$results %>% filter(!(result %in% c("PASS", "PASS WITH NOTE"))) %>% nrow(), 0)
+  expect_equal(screeningOutput$results %>% filter(!(result %in% c("PASS", "PASS WITH NOTE", "IGNORE"))) %>% nrow(), 0)
```
cjrace: What test is giving an ignore here to make you add it to the excluded list?

tests/testthat/mainTests/lsip_combinations.csv
cjrace, line 1: Can we trim this file down at all? It's noticeably bigger than our other test files 
image

R/mainTests.r
```diff
@@ -958,6 +959,17 @@ la_col_present <- function(data) {
   return(output)
 }
 
+# Function used to check if each column for that geographic level has any cells that are not blank
```
cjrace, line 962: Why is this function (col_completed) now outside of the overcomplete_regional_cols() function?

image

R/mainTests.r
```diff
@@ -967,29 +979,18 @@ overcompleted_cols <- function(data, meta) {
 
   overcomplete_regional_cols <- function(matrixRow) {
     # Start by filtering the data down to remove the geographic level being tested and any lower levels we don't care about
```
cjrace: Should we update or remove this comment now that we're explicitly stating the levels we care about? Maybe change to 'these are the levels that we would expect the region columns to always be empty for?'

R/mainTests.r
```diff
@@ -967,29 +979,18 @@ overcompleted_cols <- function(data, meta) {
...
+    levels_incompatible_with_region <- c("National", "Local skills improvement plan area")
 
     level_rows <- data %>%
-      filter(geographic_level != matrixRow[1]) %>%
-      filter(!geographic_level %in% geography_matrix[3:16, ])
+      filter(geographic_level %in% levels_incompatible_with_region)
```
cjrace: Should we still be filtering out the matrixRow level too? It was before but isn't now?

R/mainTests.r
```diff
@@ -1001,34 +1002,44 @@ overcompleted_cols <- function(data, meta) {
     # This is a test that could benefit from more detail, and maybe a table in the error feedback
 
     # Start by filtering the data down to remove the geographic level being tested, lad rows and any lower levels we don't care about
+    levels_compatible_with_la <- c("Local authority", "Local authority district", "School", "Provider", "Institution", "Planning area")
 
     level_rows <- data %>%
-      filter(geographic_level != matrixRow[1]) %>%
-      filter(!geographic_level %in% geography_matrix[13:16, ]) %>%
-      filter(geographic_level != geography_matrix[4, 1])
+      filter(!geographic_level %in% levels_compatible_with_la)
```
cjrace: Same question here as above, should we be filtering out the level that is being tested as well (matrixRow)?

R/mainTests.r
cjrace: Actually, looks like this was intentional to remove it from the scope of the function so then multiple other functions could use it and save on the copy/pasting of code?

R/mainTests.r
```diff
@@ -1120,12 +1109,15 @@ overcompleted_cols <- function(data, meta) {
...
+  lsip_matrix <- matrix(geography_matrix[7, ], nrow = 1)
 
   overcomplete_geographies <- c(
     unlist(apply(regional_matrix, 1, overcomplete_regional_cols)),
     unlist(apply(la_matrix, 1, overcomplete_la_cols)),
-    unlist(apply(geography_matrix[4:12, ], 1, overcomplete_mid_cols)),
-    unlist(apply(geography_matrix[13:16, ], 1, overcomplete_low_cols))
+    unlist(apply(geography_matrix[4:6, ], 1, overcomplete_mid_cols)),
+    unlist(apply(lsip_matrix, 1, overcomplete_lsip_cols)),
+    unlist(apply(geography_matrix[8:13, ], 1, overcomplete_mid_cols)),
```
cjrace: Could line 1117 be "...matrix[c(4:6, 8:13), ] ...." and remove the need for line 1119?

cjrace (changes requested):

R/mainTests.r
rmbielby: Yep, just code neatening. I didn't like it repeating, made the code harder to read.

tests/testthat/test-screenFiles.R
rmbielby: Ah, that's a hangover from when I was figuring out what was going on. i.e. there wasn't any LSIP data in the passes everything file, so it threw an ignore for that test. I've since added an LSIP level to the passes everything file and thought I'd removed that IGNORE in the filter. Removing it for the next commit.

R/mainTests.r
rmbielby: This is filtering it out, it's only _keeping_ levels that are in the incompatible_with_region vector.

R/mainTests.r
rmbielby: Same as above, it's being filtered out using that one filter line as it's included in the levels_compatible_with_la vector.

R/mainTests.r
rmbielby: Good point, done that.

R/mainTests.r

R/mainTests.r
rmbielby: Corrected that now

data/lsips.csv
rmbielby: Added that section to the readme

tests/testthat/mainTests/lsip_combinations.csv
rmbielby: Brought that down to the 10s of KB level now.

README.html
cjrace, line 1: Don't need to include the html for the readme - maybe one to add to git ignore?

cjrace (changes requested):

README.html
rmbielby: Done, after some fiddling around with shinytest as well.

cjrace (approved):

PR #102, opened 2023-11-17
Adding in checks for suspected wide data style column names

rmbielby:
Ethnicity and gender/sex are obvious cases where breakdowns should always be contained in either the breakdown filter or their own filter.


R/mainTests.r
```diff
@@ -2721,3 +2722,45 @@ ethnicity_characteristic_values <- function(data) {
...
+# to be 'smushed'. To do this, it flags any indicator col_name that contains
+# common filter entries (e.g. male, female, white, asian, black, etc)
+#
+# @param meta
+#
+# @return list(message, result)
+indicators_smushed <- function(meta) {
+  common_filter_substrings <- c(
+    "male", "female",
+    "white", "asian", "black", "chinese", "indian", "pakistani"
```
cjrace: Was there a reason you chose these specifically? Could we use the full table of expected filter options for ethnicity?

R/mainTests.r
```diff
@@ -2721,3 +2722,45 @@ ethnicity_characteristic_values <- function(data) {
...
+      grepl(paste(common_filter_substrings, collapse = "|"), col_name, ignore.case = TRUE)
+    ) %>%
+    pull(col_name)
+
+  if (length(indicator_names) > 0) {
+    output <- list(
+      "message" = paste0(
+        "The following indicators appear to not conform to tidy data principles: ",
+        paste(indicator_names, collapse = ", "),
+        ". We recommend pivoting your data longer and adding a filter to contain characteristic choices."
```
cjrace: Can we ask them to contact us specifically, so add a 'Please contact explore.statistics@education.gov.uk for immediate support.'?

cjrace (commented):
Couple of small questions, but generally all seems to work pretty well and doesn't seem to have broken anything else 😄

R/mainTests.r
rmbielby: I've expanded it out a little more to cover more of the range that we can filter on. Avoiding mixed and other given those could potentially be used in valid indicator names.

R/mainTests.r

cjrace:
Needs the code tidying but otherwise good to go so will approve once that test is passing @rmbielby

rmbielby:
> Needs the code tidying but otherwise good to go so will approve once that test is passing @rmbielby

D'oh!

Have run the tests now and all passing. Will add tidy code as a commit hook on the next branch I do!

cjrace (approved):

PR #106, opened 2024-01-30
Shinytest2 update

jen-machin:
Used inbuilt shinytest2 migration function to migrate tests over automatically. 
Added shinytest2 and dependencies to lockfile.
I had to modify the style of the line endings in the original test script but as this is now deleted, it should make literally no difference. The style of the line endings in the global.R file will also be modified as I saved it after adding the library call for shinytest2, but I can't imagine this causing many (if any) issues. 

Also remembered to run tidy_code ⭐ 

Can you please check that the tests all look how you were expecting and nothing has broken in the move?


cjrace:
Made some minor changes but think I need to dedicate some time to looking through the UI tests, I've got every other script running but there's a lot of changes in the snapshot I need to investigate a bit, will assign to me and pick up at a later date @jen-machin

PR #108, opened 2024-03-06
Planning area level overcompletion fix

rmbielby:
Pupil planning areas are used by the School capacity team and are used as a parent area for school level data. The screener treated them as mutually exclusive though in the `overcompled_cols()` test.

I've therefore shifted the Planning area geographic level to being a "mid" level geography (rather than a "low" level geography) in this test so that it's allowed to be a parent geographic level for school level data (and also for Provider level and Institution level).

I've tested with the latest School places data intended for publication in 2 weeks time and it now passes.

### DIFF
--- a/R/mainTests.r
+++ b/R/mainTests.r
@@ -960,12 +960,19 @@ la_col_present <- function(data) {
   return(output)
 }
 
-# Function used to check if each column for that geographic level has any cells that are not blank
-
+# Function used to check if each column for that geographic level has any cells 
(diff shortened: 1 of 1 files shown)


cjrace:
[Update how we handle planning area in the screener](https://trello.com/c/so4Zjutu/1156-update-how-we-handle-planning-area-in-the-screener)

cjrace (approved):
Happy have more comments on this function too!

PR #109, opened 2024-03-27
Add ward lookup (plus some extras)

cjrace:
## Overview of changes

This was originally to add a specific lookup for Wards, though has spiralled into a few other improvements along the way.

Currently we were just treating wards as a lower level geography and did some basic tests like avoiding duplicate names. Now we have our first team publishing this level of data on EES we've realised this isn't really suitable and instead we can pull lookups from the Open Geography Portal to get an actual list to check against.

It's worth noting that many Wards do share names, so I've included the parent LADs in the lookup file as that is what distinguishes them (along with a unique code).

1. Add a Ward lookup file

2. Generalise the existing update lookup file functions, and then use these for creating the Ward look up in a way we can keep updating it

3. Rebuild the pcon_la lookup
- I've done this as when I tested the new functions on both the existing pcon and lsip lookups, the pcon_la lookup had changes (even though I was running it with the same file that it had been built from). Digging into it I think there's a number of LADs (e.g. Barnsley) that had snook into that as LAs (e.g. cont. should be South Yorkshire) in the past. So I've switched to the version the new functions create and rebuilt it with files going back as far as 2017.
- It's worth double checking the logic where it splits off LADs from LAs, that was stolen from @rmbielby's code

4. Updated the passes_everything.csv test file
- I did this as with Ward now having its own tests, we didn't have any 'lower level geographies' in the file so some tests were now being ignore, added in MAT data to compensate and make sure every test runs and passes against that file 

5. Removed `tidy_code()` in favour of `styler::style_code()`, and updated README to match

6. Added a specific test for checking Ward combinations

7. Updated overcompleted cols test to include ward as a special case, and make the code more readable / clear on what it's doing (partly as the level of abstraction of it was tripping me up and I'm the one who originally wrote it!)

8. Changed to use the dfeR round_five_up function

EDIT:
9. Deleted tidyCode GHA, it was failing with unhelpful error messages so took the chance to move to a pre-commit hook instead as that feels like a cleaner approach anyway.

10. Also made some edits to the GitHub Actions scripts as I noticed they were using out of date actions, based this on examples from [r-lib](https://github.com/r-lib/actions/tree/v2/examples)


cjrace:
[Add Ward level lookups to the data screener](https://trello.com/c/6Ag67TdL/1166-add-ward-level-lookups-to-the-data-screener)

rmbielby:
Just noting here whilst I remember, I don't think we've got the new 2024 boundaries in yet. They're available here:
https://geoportal.statistics.gov.uk/datasets/12eff0dc5c4549a39003f73830ecbe49_0/explore

Could add as part of this PR or I'm happy to add them when I'm back after Easter.

R/standard-data-prep/utils.R
```diff
@@ -0,0 +1,168 @@
...
+      most_recent_year_included = paste0("20", new_year)
+    )
+
+  # Remove ObjectId if it exists
+  if (suppressWarnings(!is.null(new_lookup$ObjectId))) {
+    new_lookup <- new_lookup %>% select(-ObjectId)
+  }
+
+  # Extra tidy up for separating LAs / LADs
+  # Not 100% on the logic but it seems to be filtering / sifting between LAs and LADs - is that right Rich?
```
rmbielby: This bit's to deal with LAs and UTLAs not being quite the same thing. The parent UTLA for LADs in metropolitan counties (E11, e.g. Manchester) and Inner or Outer London (E13) are the Metropolitan counties and Inner / Outer London. But in these cases, the LA for our purposes is the LAD itself. 

So all this logic is doing is using the LAD as it's own parent LA if it's in a metropolitan county or in London and taking the UTLA otherwise.

E12 is the region level code, which aren't ever used as UTLAs, so doesn't actually need to be in there.

R/debugging.R
```diff
@@ -9,8 +9,8 @@ debugReadFile <- function(file) {
   return(output)
 }
 
-datafile <- debugReadFile("C:/R_Projects/dfe-published-data-qa/tests/shinytest/test-data/passes_everything.csv")
-metafile <- debugReadFile("C:/R_Projects/dfe-published-data-qa/tests/shinytest/test-data/passes_everything.meta.csv")
+datafile <- debugReadFile("C:/code_projects/dfe-published-data-qa/tests/shinytest/test-data/passes_everything.csv")
```
rmbielby: Any use of absolute paths makes me nervous! Can kind of see why you might want it to be absolute here, but would still be tempted to shift it to relative given it is just pointing to file in the repo in this case.

.hooks/pre-commit.R
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env Rscript
+cat("\nRunning commit hooks...", fill = TRUE)
```
rmbielby: Any particular reason for using `cat` rather than `message`?

R/mainTests.r
```diff
@@ -983,68 +984,31 @@ col_completed <- function(x, row, level_rows) {
...
-
-    # Start by removing any geographic levels that are allowed to have entries in the LA ID and name columns
-    levels_compatible_with_la <- c("Local authority", "Local authority district", "School", "Provider", "Institution", "Planning area")
-
-    level_rows <- data %>% filter(!geographic_level %in% levels_compatible_with_la)
+  # Helper function for calculating overcomplete rows
+  # Feed in the geog level you want to test, and then the levels it's columns it is okay to be completed for
+  # It will return any columns that are detected as filled in for rows that they shouldn't be
+  #
+  # E.g. if you use geog_row = 4, that is the LAD row in the geog matrix...
```
rmbielby, line 991: Do feel like it'd be good to switch to a data frame rather than a matrix at some point. Suspect it could make it more readable to use filter() and select() etc rather than using the matrix indices. Don't think it's worth the hassle adding it to a big PR like this, but would be tempted to look at trying it in a standalone PR.

rmbielby (commented):
Mainly just comments rather than changes so far.

cjrace:
> Just noting here whilst I remember, I don't think we've got the new 2024 boundaries in yet. They're available here: https://geoportal.statistics.gov.uk/datasets/12eff0dc5c4549a39003f73830ecbe49_0/explore
> 
> Could add as part of this PR or I'm happy to add them when I'm back after Easter.

Added this to trello to pick up separately. I think it'll need to be a new lookup that's just for PCons as the 2025 boundaries don't have the LA lookup yet - https://trello.com/c/Fz9Hlyeu/1288-add-pcon-boundary-file-to-the-screener-specifically-with-2025-boundaries-in

R/standard-data-prep/utils.R
cjrace: Nice, thanks! Have stolen this and updated the comment to explain the logic, and then have removed E12 from it

R/debugging.R
cjrace: Fair point! These have been like this for ages, to be honest I can just remove entirely and leave the `file.choose()` lines in as that's what I use most when debugging given the offending file is often in a downloads or other folder.

Updated this to remove the absolute paths

.hooks/pre-commit.R
cjrace: No reason, think it's what you used and I just copied it to here! So right back at you - any reason for one over the other?

R/mainTests.r
cjrace: Yeah I think that's outside of this PR. There's some more of the code that would need updating as the current code needs it in a matrix to easily apply the function over the rows and columns (the reason why I used one to start with as it was the quickest way I could figure out to do what I want).

It's possible to do frame a dataframe but would need more rewriting than I'd like to do here.

.hooks/pre-commit.R
rmbielby: Hah, no I must have just written that code you used before I took the step of a conscious decoupling from `cat`, and used `message` everywhere instead.

R/mainTests.r
rmbielby: That's what I figured. I'll add it to the issues log in case we do want to pick it up at some point.

R/standard-data-prep/utils.R
```diff
@@ -0,0 +1,177 @@
...
+  # This bit's to deal with LAs and UTLAs not being quite the same thing. 
+  # The parent UTLA for LADs in metropolitan counties (E11, e.g. Manchester),
+  # and Inner or Outer London (E13) are the Metropolitan counties 
+  # and Inner / Outer London. 
+  # But in these cases, the LA for our purposes is the LAD itself.
+  
+  # So this is using the LAD as it's own parent LA if it's in a metropolitan
+  # county or in London and taking the UTLA otherwise.
+  
+  So all this logic is doing is using the LAD as it's own parent LA if it's in a metropolitan county or in London and taking the UTLA otherwise.
```
rmbielby: Missed a # here

R/standard-data-prep/utils.R
```diff
@@ -0,0 +1,177 @@
...
+
+  # Append the new lookup
+  updated_lookup <- existing_lookup %>%
+    rbind(new_lookup) %>%
+    # Then condense the rows, rewriting the first and last years for each row
+    summarise(
+      first_available_year_included = min(first_available_year_included),
+      most_recent_year_included = max(most_recent_year_included),
+      .by = all_of(cols_to_join_by)
+    )
```
rmbielby, line 197: I'd quite like a bit of clean up here using `arrange()`.

I'm thinking maybe something along the lines of:
```
%>% arrange(names(new_lookup)[names(new_lookup) %in% c("ward_code", "pcon_code", "lad_code", "new_la_code", "lsip_code")])
```

data/ward_lad_hierarchy.csv
rmbielby, line 1: I had a bit of a look for duplicates in this file and there's a handful that I think we might want to clean up:

- Some Wards switch from being "St ?????" to "St. ?????", could do a gsub to convert all "St " to "St. "?
- Some Welsh wards have some sort of accent over a "y" which hasn't been encoded properly in some of the files so comes out as a "?": "Pen draw Llŷn", "Rhyl Tŷ Newydd"
- Littledown and Iford has a typo in the 2019 ONS data and comes up as Littledown and Ilford instead

Everything else looks like a genuine name change.

One extra random question that comes to mind looking through the data, is whether we want to standardise "and"?! The source files are inconsistent between using "&" or "and" in names (e.g. "Jacksdale & Westwood" versus "Marple South and High Lane"). I know we try and avoid symbols, but not sure if we'd rather the lookup was consistent with ONS names over being internally consistent.

data/la_pcon_hierarchy.csv
rmbielby, line 1: One annoying little duplicate in this one, which I've seen trip teams up before:
"Birmingham,  Selly Oak" in 2019 and 2020, instead of "Birmingham, Selly Oak"

rmbielby (changes requested):
Few more comments. Gradually getting more thorough with this one!

.hooks/pre-commit.R
cjrace: Fair enough - will update to `message()`

R/standard-data-prep/utils.R
cjrace: Huh weird, no idea how the commit hooks didn't catch that, they catch any failing scripts and this script would fail (plus just tried another commit this morning and it caught it). Fixed in new commits

data/ward_lad_hierarchy.csv
cjrace: I'm torn on this. Part of me agrees and thinks we should do the gsub, and standards for 'and'. Though another part thinks we should be faithful to the ONS codes over the years, and treat them all as legitimate, rather than setting a higher bar than ONS themselves.

I think we either 1) just fix the encoding issues, or 2) fix absolutely everything, likely by starting a variations or fixes script that contains and executes all our additional rules. I'd lean towards the former for simplicity on our part, though interested to know more of your thoughts, as I could easily be swayed to the latter?

data/la_pcon_hierarchy.csv
cjrace: This took a ridiculous amount of time to realise what you meant - found the double spacing now though! I think it ties in with the question above. If we're taking a minimal approach and just showing what ONS have then I don't think we fix this.

On the other hand, if we did decide to clean up, then this could be a rule we add that only ever allows a single space within a place name and strips out any excess automatically.

R/standard-data-prep/utils.R
cjrace: What's the clean up you'd want to see?

I'm not sure we'd want to be as specific in this function? My thinking was this was fairly generalised so that any columns could be passed through and then if you wanted to filter any out, that's easy enough to do before you run the 'write' function?

data/ward_lad_hierarchy.csv
rmbielby: I'd throw the St. / St variations in with Welsh accent encoding issue as something to fix and leave the & / and as is. For Ilford (and the St. variation), I might drop ONS a message and see if they're interested in fixing that one on their data files.

data/la_pcon_hierarchy.csv
rmbielby: I think my angle is the point above that I've seen exactly this item cause issues  in team's data (and like you say, it can be a hard one to spot). So by adding this in the current form, we're providing reference data that could potentially cause team's issues. I guess this is minimised as we'd generally expect teams only to use the pcons from the latest year, but I guess we've got a chance to put in prevention for if it were to happen again.

R/standard-data-prep/utils.R
cjrace: We spoke about this, will add some generic ordering so it's done by year available and the code column

data/ward_lad_hierarchy.csv
cjrace: If we're going to ask ONS then I'd be tempted to leave out the St. variations for now and just fix the encoding issue, does that work as a plan, and then we can tidy up further in a future PR if we want to based on ONS's response. 

Is it worth adding in the question to them about whether they have a standard they'd want to apply to & versus 'and'?

data/la_pcon_hierarchy.csv
cjrace: Makes sense, happy to add this one in

R/standard-data-prep/utils.R
cjrace: Added now EDIT: [might not be working, checking this myself again locally]

data/la_pcon_hierarchy.csv
cjrace: Added now

cjrace:
> 3. Rebuild the pcon_la lookup
> 
> * I've done this as when I tested the new functions on both the existing pcon and lsip lookups, the pcon_la lookup had changes (even though I was running it with the same file that it had been built from). Digging into it I think there's a number of LADs (e.g. Barnsley) that had snook into that as LAs (e.g. cont. should be South Yorkshire) in the past. So I've switched to the version the new functions create and rebuilt it with files going back as far as 2017.

Worked out what was happening here, was actually me not realising that Barnsley sits in a metropolitan county, so is both the LAD and LA, even when South Yorkshire is the UTLA. It's been an education!

data/ward_lad_hierarchy.csv
rmbielby: Fair enough. Just the Welsh fix for now.

cjrace: Knew I'd forgotten something... Will fix that and let you know once it's in

cjrace: Have had a go @rmbielby, but I think there's some encoding in names left that are still an issue. However, these are wrong at source, so there's nothing I we can do unless we do a manual fix for them.

I think we lump into the list of things to talk to ONS about and see if they'll fix the source data instead?

Seems to be specifically the 2022 file that has the issues, it's fine in the 2023 one.

https://geoportal.statistics.gov.uk/datasets/bc7706efd6e4413a8d5113af23f16d77_0/explore?filters=eyJXRDIyTk0iOlsiUmh5bCBUPyBOZXd5ZGQiXX0%3D

image

rmbielby: Adding the update from ONS for completeness - they've fixed Birmingham, Selly Oak on their side and Littledown and Iford, but won't touch any of the rest due to being constrained by Statutory Instruments.

rmbielby (dismissed):

cjrace:
@rmbielby - I have updated the tests to use shinytest2 now as we spoke about on this branch, done a few bits of tidy up around that too. Tagging @jen-machin for sight too as if we're happy with the changes here I can close #106.

The tests themselves should generally be much clearer / easier to see genuine failures now too as I've taken advantage of exporting values from the app in places where before we had masses of html output to sift through.

All of the changes are in a single commit so you can review the diff there. The only exception is some workflow faffing afterwards that ended up a bit trial and error eske. Happy to chat through if you have any questions!

rmbielby:
> @rmbielby - I have updated the tests to use shinytest2 now as we spoke about on this branch, done a few bits of tidy up around that too. Tagging @jen-machin for sight too as if we're happy with the changes here I can close #106.
> 
> The tests themselves should generally be much clearer / easier to see genuine failures now too as I've taken advantage of exporting values from the app in places where before we had masses of html output to sift through.
> 
> All of the changes are in a single commit so you can review the diff there. The only exception is some workflow faffing afterwards that ended up a bit trial and error eske. Happy to chat through if you have any questions!

image

rmbielby (approved):
Checks all pass on a run here. 

I've gone through and manually followed some of the shiny test steps to make sure what the tests are producing is what I'd expect given the inputs.

PR #110, opened 2024-04-16
Adding lookups for standardised school types and SEN filters

rmbielby:
Just adding the data reference files for SEN and school type categories.

I haven't added any actual checks around these yet, but figure it's useful to have them available to users whilst we're developing the screening checks.


data/school-types.csv
```diff
@@ -0,0 +1,72 @@
...
+establishment_type,Establishment type,Academy sponsor led,education_phase;establishment_type_group;establishment_governance,,
+establishment_type,Establishment type,All establishments,education_phase;establishment_type_group;establishment_governance,,
+establishment_type,Establishment type,City technology college,education_phase;establishment_type_group;establishment_governance,,
+establishment_type,Establishment type,Community school,education_phase;establishment_type_group;establishment_governance,,
+establishment_type,Establishment type,Foundation school,education_phase;establishment_type_group;establishment_governance,,
+establishment_type,Establishment type,Free school,education_phase;establishment_type_group;establishment_governance,,
+establishment_type,Establishment type,Free school 16 to 19,education_phase;establishment_type_group;establishment_governance,,
+establishment_type,Establishment type,Hospital special school,education_phase;establishment_type_group;establishment_governance,,
+establishment_type,Establishment type,Other FE sector college,education_phase;establishment_type_group;establishment_governance,,
```
cjrace, line 41: What do the multiple options here mean? Is that saying there's there's different available options, or that at some point we'd expect to have all of these as layers in the same hierarchy?

data/school-types.csv
```diff
@@ -0,0 +1,72 @@
+col_name,breakdown_topic,filter_item,filter_grouping_column,filter_item_grouping,school_census_code
+establishment_type_group,Establishment type group,Academy,,,
+establishment_type_group,Establishment type group,Total,,,
+establishment_type_group,Establishment type group,All state funded,,,
+establishment_type_group,Establishment type group,Independent,,,
+establishment_type_group,Establishment type group,LA maintained,,,
```
cjrace: Did you consider using 'Local authority maintained' instead of just LA? (I think we can argue it either way)

data/sen.csv
```diff
@@ -0,0 +1,42 @@
+col_name,breakdown _topic,filter_item,filter_grouping_column,filter_item_grouping,school_census_code
+sen_status,SEN status,Total,,,
+sen_status,SEN status,Any special educational need,,,
+sen_status,SEN status,No identified special educational need,,,
+sen_provision,SEN provision,"Education, health and care plan",sen_status,Any special educational needs,
```
cjrace: Is there a reason for having plural items for sen_status when grouping but in the first 3 rows we use singular versions of the same phrase? Would it be easier to pick one to help with matching data?

data/school-types.csv
rmbielby: Those are to denote multiple possible choices for grouping column (given the current 2-tier hierarchy) that teams might have in their data.

data/school-types.csv
rmbielby: LA maintained is just the current school census standard. This could be a good chance to push towards avoiding the acronym to be fair.

data/sen.csv
rmbielby: Nope, that's a typo, good spot.

data/school-types.csv
cjrace: Might as well take the chance to push for it then! Do you want to do that before merging this, or merge this and then push in parallel and update here if we change it later?

cjrace (dismissed):

data/school-types.csv
rmbielby: I've updated away from he acronym here and on the guidance. Ann didn't have any problem with expanding it out.

cjrace (approved):

PR #111, opened 2024-05-03
Meta label missing bug

cjrace:
# Brief overview of changes

Addressed a bug where a metadata CSV could have a missing label for a filter / indicator but still pass the screening.

Missing metadata labels are important to catch as they will prevent files from being uploaded to EES (and because all filters / indicators should have human readable versions for use in the service).

Added example test file and test case under 'edge cases', and then fixed the offending function by accounting for blanks as well as NAs.

1. No idea how this hasn't come up before now, though a good one to have found and fixed.

2. Looking at the example file we had reported to us (left), you can see if opening in notepad that each cell is wrapped in quotes, which differs from the `label.csv` example we had in the main tests, an equivalent of this is on the right of the image.

image

3. I also updated the debugging script as I realised after my recent PR #109, we needed to source more dependencies for the debugging to work (as we make use of Shiny autoloading scripts from the R folder for normal app running, so the global script no longer sources all dependencies alone).

4. The final UI test for the QA tabs was also failing intermittently on me, adding an extra wait seems to have made it pass reliably, so hoping this will also fix the existing failure in the GitHub actions on the master branch.

### DIFF
--- a/R/mainTests.r
+++ b/R/mainTests.r
@@ -1959,6 +1959,7 @@ col_name_spaces <- function(meta) {
 label <- function(meta) {
   completed_labels <- meta %>%
     filter(!is.na(label)) %>%
+    filter(label != "") %>%
     nrow()
 
   blank_labels <- nrow(meta) - completed_labels
--- a/R/manual_scripts/debugging.R
+++ b/R/manual_scripts/debugging.R
@@ -1,7 +1,18 @@
 # Debugging script
 # Use this script to manually read in and screen files for those occasions when the app breaks without a useful error
+# Helpful for running code in console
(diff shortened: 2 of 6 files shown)


cjrace:
[Fix blank label bug in screener](https://trello.com/c/owezW738/1342-fix-blank-label-bug-in-screener)

rmbielby (approved):
Had a look through the changes, ran the tests and ran the screener with the new test case demo file. All seemed good.

Am assuming it's `data.tables::fread()` that's changed behaviour. Had a quick look through their NEWS section - plenty of changes around strings during read-in, but no specific reference to empty cells being read as empty strings rather than NAs.

PR #113, opened 2024-05-16
Added Outside of the United Kingdom and unknown to geographies

rmbielby:
HE Graduate Outcomes have been using "Outside of England and unknown", but have now concluded that they need "Outside of the United Kingdom and unknown" instead. I've added it as an option to all reference data that already had "Outside of England and unknown", which covers:

- Country
- Region
- Local authorities
- Local authority districts
- Parliamentary constituencies

"Outside of England and unknown" still seems like a valid item as well (and I think was originally requested by FE and Skills), so keeping them as alternative options to each other depending on what the data in a given publication covers.


global.R
cjrace, line 1: Should knock this back down @rmbielby to protect the servers until we have more control of our own?

cjrace (changes requested):

global.R
rmbielby: Put it back how I found it now!

cjrace (approved):

PR #114, opened 2024-05-21
add 2024 pcon to postcode lookup

cjrace:
Quick one, Catherine asked us to add in the latest PCon boundaries and the postcode look up for teams to use.

File exists in the data/ folder to be used as a lookup.

Existing tests for PCon's extended to include all new locations from this file too.

PR #115, opened 2024-05-28
remove postcode lookup

cjrace:
Have removed the postcode lookup as added in #114, as that was only half the file, the full file is too big. Instead we'll just point teams to the ONS website to download from directly.

To compensate as far as the screening checks are concerned, I've pulled in the latest constituencies as a separate file until ONS updates the main Ward > PCon > LAD > LA lookup that we use elsewhere.

### DIFF
--- a/R/knownVariables.r
+++ b/R/knownVariables.r
@@ -83,11 +83,10 @@ lads <- suppressMessages(read_csv("data/lads.csv"))
 leps <- suppressMessages(read_csv("data/leps.csv"))
 edas <- suppressMessages(read_csv("data/english_devolved_areas.csv"))
 
-# PCons, this is quick and dirty, should take the chance to clean up when we move to using Pins
+# PCons, this is quick and dirty, should move to use the ward / PCon / LAD / LA file instead
 pcons <- suppressMessages(read_csv("data/pcons.csv"))
 pcons_2024 <- suppressMessages(read_csv("data/pcon_2024_v2.csv")) %>%
-  rename("pcon_code" = PCON24CD, "pcon_name" = PCON24NM) %>%
-  select(c(pcon_code, pcon_name))
+  select(pcon_code = PCON24CD, pcon_name = PCON24NM)
 
 combined_pcons <- rbind(pcons %>% select(pcon_code, pcon_name), pcons_2024)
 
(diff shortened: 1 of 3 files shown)


cjrace:
Adding @rmbielby to review as this is effectively undoing what we did in #114

cjrace:
[Remove PCon postcode lookup from screener](https://trello.com/c/RcsSrKHL/1386-remove-pcon-postcode-lookup-from-screener)

R/knownVariables.r
```diff
@@ -85,9 +85,9 @@ edas <- suppressMessages(read_csv("data/english_devolved_areas.csv"))
 
 # PCons, this is quick and dirty, should take the chance to clean up when we move to using Pins
 pcons <- suppressMessages(read_csv("data/pcons.csv"))
-pcons_2024 <- suppressMessages(read_csv("data/postcode_to_pcon_2024.csv")) %>%
-  distinct(pconcd, pconnm) %>%
-  rename("pcon_code" = pconcd, "pcon_name" = pconnm)
+pcons_2024 <- suppressMessages(read_csv("data/pcon_2024_v2.csv")) %>%
+  rename("pcon_code" = PCON24CD, "pcon_name" = PCON24NM) %>%
```
rmbielby: You could just do this as one single line:
`select(pcon_code = PCON24CD, pcon_name = PCON24NM)`

data/pcon_2024_v2.csv
rmbielby, line 1: Any particular reason for the "_v2" suffix on this filename?

data/pcon_2024_v2.csv
```diff
@@ -0,0 +1,651 @@
+﻿PCON24CD,PCON24NM,PCON24NMW,ObjectId
```
rmbielby, line 1: Is there any reason to keep the ObjectId field?

rmbielby (commented):
Few thoughts from looking through the files:
- Are we going to just merge pcons_2024_v2.csv into pcons.csv at some point and apply a date of termination / date of introduction so we're consistent in how we deal with geographies and have only one base pcon file?
- If we're keeping both for now, do we want to rename the pcons.csv file to have a suffix like pre-2024?

data/pcon_2024_v2.csv
cjrace: It's what ONS had called it

data/pcon_2024_v2.csv
cjrace: Just to keep faithful to the original file from ONS, so we've not transformed it in anyway (and to be lazy...)

R/knownVariables.r
cjrace: Nifty! Have updated to this in my PR response commit

rmbielby (approved):

PR #120, opened 2024-07-06
Incorrect provider cols bug

cjrace:
EDITED!

# Brief overview of changes

Fix the expected geography cols check to cover provider and institution level. 

Big refactor to make code using the geography_matrix (levels v columns) more readable.

We found the app crashes if the expected provider columns aren't present. Those expected columns are required in the main checks and previously we'd failed files like this in pre_checks_2.

We suspect this happened in a previous change when editing the geography_matrix. It highlights the current risks around the level of abstraction in the code relating to geographic levels and columns and how hard it can be to make safe changes to them. By refactoring the code to be more explicit and readable it should be easier to maintain and make future bugs less likely. 

Original
1. Added example test data
2. Added edge case test
3. Fixed the geography_present within the pre_check_2 script to prevent files like this getting to the main tests where the app will crash.

Additional

1. Fixed the geography_present within the pre_check_2 script to prevent files like this getting to the main tests where the app will crash.
2. Moved the geography_matrix to a CSV file. 
3. Replaced uses of geography_matrix, with geography_dataframe or hard coded variables where easy to do so, leaving the matrix usage only where it is helpful for applying functions over.
4. Added a script of unit tests, that check the assumed properties of the geography matrix / dataframe. This will help to highlight if any assumptions are invalidated in future changes.
5. geography_level_completed - this test previously only checked up to row 14 (school), though in theory should check all but planning area, so have changed the scope of this test to match. This should also 
6. Fix the variable_snake_case() check. When writing the unit tests for the matrix I realised that the check wasn't checking for spaces in variable names. I know these are checked earlier, though for completeness of the function I've adapted it so it will capture any spaces too.
7. Updated the email address in the main UI to point to the EES platforms team.
8. Update an error message that previously pointed to SDT to point to the EES platforms team.

Original bug created in PR #98, created by rmbielby and reviewed by cjrace, and if those as the pair who have written the vast majority of this app can miss this, then anyone can.

The refactor does make a more towards more hardcoded references to geographic levels and associated column names, which may cost time should we change the labels specified in the guidance in future. Though given there already were some in the code as it was, it seems appropriate to consciously move towards more explicit and readable code to help the maintainability of the app and functions long term and aim to reduce the likelihood of bugs like this as a higher priority.

The overcompleted_cols() check is still a bit of a beast, though continues to work and the other tests aren't paying as much of a price anymore at least!

No linked issues. Links to https://dfedigital.atlassian.net/browse/EES-5319 and was found when screening those files.


cjrace:
[Fix bug in screener crashing with old 'dodgy' EES files](https://trello.com/c/vMIOL9oX/1489-fix-bug-in-screener-crashing-with-old-dodgy-ees-files)

rmbielby (commented):
I suspect this is a hangover from the LSIP add I did. Main question I have is if you checked all the other references to `geography_matrix` in this script? My guess is that I missed this script in the LSIP update and so nothing's been updated.

Main line I'm suspicious of is:
`acceptable_ob_units_sch_prov_filter <- acceptable_observational_units[!acceptable_observational_units %in% c(geography_matrix[13, 3], geography_matrix[14, 3])]`

The variable being created looks like it's doing something to filter for school and provider, but the entries in the matrix on rows 13 and 14 are Sponsor and School rather than School and Provider.

cjrace:
> I suspect this is a hangover from the LSIP add I did. Main question I have is if you checked all the other references to `geography_matrix` in this script? My guess is that I missed this script in the LSIP update and so nothing's been updated.
> 
> Main line I'm suspicious of is: `acceptable_ob_units_sch_prov_filter <- acceptable_observational_units[!acceptable_observational_units %in% c(geography_matrix[13, 3], geography_matrix[14, 3])]`
> 
> The variable being created looks like it's doing something to filter for school and provider, but the entries in the matrix on rows 13 and 14 are Sponsor and School rather than School and Provider.

Have made a big update to this now - including adding a lot to the original PR description. Worth re-reading that again on re-review @rmbielby

R/mainTests.r
```diff
@@ -752,35 +760,35 @@ region_for_la <- function(data) {
 # check if there is LA level data, and if so, if regional columns are present and completed
```
rmbielby: Small typo in comments - LA -> LAD

rmbielby: Having said that, I feel like this would be a prime one to consolidate into a single function at some point (not now):

`region_present <- function(data, level = 'Local auhourity'){...}`

data/geography_matrix.csv
```diff
@@ -0,0 +1,18 @@
+row_number,geographic_level,code_field,name_field,code_field_secondary
```
rmbielby, line 1: Aware this has been a big rewrite, but just thinking around a middle ground between avoiding hard-wiring and getting the readability / reliability. I'm thinking of adding a column to the dataframe called level_shorthand, which is our unchanging internal screener code for each geographic level. Then we'd have a function to pull info out:
```
geog_info <- function(level, param){
   geographic_dataframe %>% 
      filter(level_shorthand == level) %>%
      pull(param)
}
```
And then everywhere you've hardwired "region_code", you can replace that with `geog_info("region", "code_field")` or `geog_info("la", "name_field")` for "la_name" or `geog_info("lad", "geographic_level")` for "Local authority district". 

It's short, doesn't rely on anonymous numerics and allows for us randomly needing to rename a whole geographic level in EES (which I guess is low probability, but feels like something you were trying to guard against previously).

Feel free to tell me sod off.

rmbielby (commented):
Had a thought on how to adapt this to be potentially a bit more future proof.

R/mainTests.r
cjrace: Fixed the typo and added #126 as you're quite right, they could very ees-ily be combined!

data/geography_matrix.csv
cjrace: Oooh. I like this suggestion a lot. Though, I will politely tell you to sod off for this PR at least and suggest we do this at a later point.

I've raised #127 for this, as it would be nice to move to this. Though when I did the rewrite in this PR I looked for all references to the geography_matrix, if you look through for any references to geographic locations / columns there's a whole raft more of hard coded stuff, which is partly what chipped away at my attempts to guard against it as I realised I was trying to hold back the floodgates while a river was happily also running past me!

rmbielby (approved):
Looking good

PR #122, opened 2024-07-06
remove 3 year timeseries test

cjrace:
This test was originally added at the start as we were trying to encourage all teams to have at least 3 years of data in EES to cover teams back to 2018 when the accessibility regulations came into force. The theory was, if they included a time-series of at least 3 years in a file, they were covering back to at least 2018.

This is now redundant (and has been for a while). To reduce excess code / processing, and also the risk of warning fatigue / blindness, I've now removed it.

Bonus:

I've also added caching back for the packages used in renv for the automated tests GitHub action. The path given doesn't work so doesn't save a cache, but seems to work well enough to allow the caching step to happen and it seems to find a general cache for that version of linux / R. Not entirely sure how or why, though it cut the time from 5 mins to 2 mins so seems to do the trick 🤷

### DIFF
--- a/.github/workflows/automatedTests.yaml
+++ b/.github/workflows/automatedTests.yaml
@@ -24,6 +24,14 @@ jobs:
           
       - name: Install git2r dependencies
         run: sudo apt-get install -y libgit2-dev
+
+      - name: Cache renv packages
+        uses: actions/cache@v4
+        with:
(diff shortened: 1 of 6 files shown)


cjrace:
[Remove 3 years data check from screener](https://trello.com/c/BDiwBbH9/1490-remove-3-years-data-check-from-screener)

rmbielby (approved):

PR #129, opened 2024-08-15
Update guidance URLs

cjrace:
# Brief overview of changes

We give guidance links in a number of messages on our tests, have updated those to the analysts guide URL

The rsconnect site hasn't been updated for years and was binned altogether earlier this year / last year (a while ago!).

Mass find and replace of the domain name for the new domain, all pages and sections have similar titles now so didn't change those.

None

None

### DIFF
--- a/R/fileValidation.r
+++ b/R/fileValidation.r
@@ -189,18 +189,18 @@ file_separator <- function(data, meta) {
     if ((grepl(",", data) == FALSE) && (grepl(",", meta) == FALSE)) {
       output <- list(
         "result" = "FAIL",
-        "message" = paste0("Neither of the data and metadata files appear to be using a comma delimited structure. <br> - See the ", "<a href='https://dfe-analytical-services.github.io/analysts-guide/statistics-production/ud.html#data_format' target='_blank'>guidance on how to save the files</a>", " if you are unsure.")
+        "message" = paste0("Neither of the data and metadata files appear to be using a comma delimited structure. <br> - See the ", "<a href='https://dfe-analytical-services.github.io/analysts-guide/statistics-production/ud.html#data-format' target='_blank'>guidance on data formats</a> and <a href='https://dfe-analytical-services.github.io/analysts-guide/statistics-production/ud.html#how-to-export-data-with-utf-8-encoding' target='_blank'>guidance how to save data files in the right format</a>", " if you are unsure.")
       )
     } else {
       if (grepl(",", meta)) {
         output <- list(
           "result" = "FAIL",
-          "message" = paste0("The data file does not appear to be using a comma delimited structure. <br> - See the ", "<a href='https://dfe-analytical-services.github.io/analysts-guide/statistics-production/ud.html#data_format' target='_blank'>guidance on how to save the files</a>", " if you are unsure.")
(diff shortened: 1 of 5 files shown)


R/fileValidation.r
```diff
@@ -189,18 +189,18 @@ file_separator <- function(data, meta) {
     if ((grepl(",", data) == FALSE) && (grepl(",", meta) == FALSE)) {
       output <- list(
         "result" = "FAIL",
-        "message" = paste0("Neither of the data and metadata files appear to be using a comma delimited structure. <br> - See the ", "<a href='https://rsconnect/rsc/stats-production-guidance/ud.html#data_format' target='_blank'>guidance on how to save the files</a>", " if you are unsure.")
+        "message" = paste0("Neither of the data and metadata files appear to be using a comma delimited structure. <br> - See the ", "<a href='https://dfe-analytical-services.github.io/analysts-guide/statistics-production/ud.html#data_format' target='_blank'>guidance on how to save the files</a>", " if you are unsure.")
```
jen-machin: I don't know whether this is a quirk of the browser I'm using but this link takes me to the top of the Open Data Standards page and not the relevant section - not the end of the world but slightly irritating

R/fileValidation.r
```diff
@@ -189,18 +189,18 @@ file_separator <- function(data, meta) {
...
       )
     } else {
       if (grepl(",", meta)) {
         output <- list(
           "result" = "FAIL",
-          "message" = paste0("The data file does not appear to be using a comma delimited structure. <br> - See the ", "<a href='https://rsconnect/rsc/stats-production-guidance/ud.html#data_format' target='_blank'>guidance on how to save the files</a>", " if you are unsure.")
+          "message" = paste0("The data file does not appear to be using a comma delimited structure. <br> - See the ", "<a href='https://dfe-analytical-services.github.io/analysts-guide/statistics-production/ud.html#data_format' target='_blank'>guidance on how to save the files</a>", " if you are unsure.")
```
jen-machin: Somewhat outside the scope of this PR but I don't feel like "comma delimited structure" is well enough explained in the analysts guide page - I know what it means but not sure everyone will and I think we could do with an actual example of what different delimiters are and how they're used. Doesn't have to be excessive but could help people spot incorrect formats more quickly if they're unfamiliar with the concept

R/fileValidation.r
cjrace: Ah damn, this is my laziness coming through - the anchor tag has changed to use kebab case in quarto so should be `#data-format`, I'll update and double check the others too

R/fileValidation.r
cjrace: That's fair, for this I've updated it to link to both that section and the part for saving as UTF-8 as that at least gives instructions on how to save files in the right format. Also raised https://github.com/dfe-analytical-services/analysts-guide/issues/69 on the analysts guide for someone to pick up separately to this PR

R/fileValidation.r
cjrace: Fixed this in the latest commit

jen-machin:
[Update guidance URLs on data screener](https://trello.com/c/9AInmdI7/1684-update-guidance-urls-on-data-screener)

R/fileValidation.r
jen-machin: I got caught out by this yesterday too!

jen-machin (approved):

PR #131, opened 2024-08-20
Fix bug with missing region name

cjrace:
# Brief overview of changes

Fixed a bug that happened in files with only one region column present and no region rows.

The app currently will error and then crash with no feedback to users, we want to do better than that and always give feedback on failing files.

Example when running the app locally on the main branch with the test files (same error in the prod logs) 
image

- Added a test file that errors on the production site currently.
- Added a unit test to check there is no error when screening that file (this failed until I added the fix, and now it passes)
- Updated the `region_combinations()` test to only execute if both region_name and region_code columns are present
- Updated the `country_combinations()` test to remove the check for country_code being present as no file can get to the main checks stage without it

The test I've fixed - `region_combinations()` doesn't need to fail if only one region column is present as another test `region_col_present()` checks for this already.

To make it easy to test, you can use the test data files I've added in this PR, try screening them against the current prod version - https://rsconnect/rsc/dfe-published-data-qa/, and then try doing the same on the dev environment (where this has deployed to after I opened the PR - https://rsconnect-pp/rsc/dev-dfe-published-data-qa/). You should see the app on prod will crash but it'll handle it neatly on the dev environment.

No related issues

### DIFF
--- a/R/mainTests.r
+++ b/R/mainTests.r
@@ -1493,11 +1493,12 @@ ward_combinations <- function(data) {
 # region_combinations -------------------------------------
 # Checking that region_code and region_name combinations are valid
 # We know from geography_level_present (pre-check 2) that if regional rows exist both cols must be present
+# We know from region_col_present (main tests) that if one region col exists both cols must be present
 
 region_combinations <- function(data) {
-  if (!"region_code" %in% names(data)) {
+  if (!"region_code" %in% names(data) || !"region_name" %in% names(data)) {
     output <- list(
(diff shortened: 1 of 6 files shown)


cjrace:
[Error causing data screener to fall over](https://trello.com/c/P0e1e5F5/1695-error-causing-data-screener-to-fall-over)

jen-machin (approved):
This works as expected on the dev version, definitely less frustrating than the old version!

PR #133, opened 2024-08-27
Standardised col_name check

rmbielby:
# Brief overview of changes

I've expanded the ethnicity headers check to allow checks of any standardised col_name that comes under harmonisation rules. This runs off a look-up table of common aliases or sub-strings used in publications that we can check on and their related standard col_names. The test does a scan for these aliases / sub-strings and throws an error if it finds them as non-standard col_names.

End users currently have a confusing collection of different col_names across different data files. We'd like to improve consistency in the data catalogue.

The initial look-up table is provided in data/harmonised_col_names.csv. This can be added to as more themes fall under the standardised bracket. For now we have:
- ethnicity
- SEN
- Establishment types and phases

The `ethnicity_headers()` function has been made more generic and renamed as `standard_filter_headers()`.

Lots of the test data failed the more generalised test, so I've gone through and updated all the data files that were now failing to meet the latest standards.

I've not added any tests for standardised filter items as part of this test as it was already breaking lots of the unit tests as it was, so has become quite a substantial update despite it being one small test that's been updated. Feel like this is about right for a single PR and I can add further tests on a second PR.


cjrace:
[data screener - standardised col_name check](https://trello.com/c/UXD1QcAV/1702-data-screener-standardised-colname-check)

R/mainTests.r
```diff
@@ -2556,36 +2556,39 @@ indicator_dp_completed <- function(meta) {
...
       "message" = paste0(
-        "The following columns appear to relate to ethnicity data, but do not conform to the standard col_name conventions: <br> - '",
-        paste(ethnicity_columns, collapse = "', '"), "'. <br> - These should take the form of one of the following: ",
-        paste(ethnicity_standard_headers, collapse = ", "),
-        "."
+        "The column(s) ",
+        paste(bad_col_names, collapse = "', '"), " appear to relate to ",
+        "contexts that fall under the harmonised data standards. Please verify ",
+        "your column headers against the data standards in:\n",
+        "https://github.com/dfe-analytical-services/dfe-published-data-qa/tree",
```
cjrace: Can we make this a clickable URL? 

E.g. https://github.com/dfe-analytical-services/dfe-published-data-qa/blob/a9570b1248b64f6eb52c3be6f51e92f6e754be60/R/mainTests.r#L144

Though use something 'DfE harmonised data standards' for link text

cjrace: Should the link be to the analysts guide rather than the screener repo?

data/harmonised_col_names.csv
```diff
@@ -0,0 +1,17 @@
+col_name_search_string,col_name_harmonised
+sen_,sen_status
```
cjrace, line 2: is there a reason for sen having an underscore after it?

global.R
```diff
@@ -1,5 +1,5 @@
 # Max file size ---------------------------------------------------------------------------------
-options(shiny.maxRequestSize = 800 * 1024^2)
+options(shiny.maxRequestSize = 1200 * 1024^2)
```
cjrace: Intentional change or accidental? You didn't reference this in the PR description

tests/testthat/mainTests/ethnicity_meets_standards.csv
cjrace, line 1: Could this file be smaller? (would also make reviewing in Git easier)

R/mainTests.r
```diff
@@ -2556,36 +2556,39 @@ indicator_dp_completed <- function(meta) {
...
-  ethnicity_columns <- meta %>%
-    filter(
-      grepl("ethnic", tolower(col_name)),
-      !(col_name %in% ethnicity_standard_headers)
-    ) %>%
-    pull(col_name)
-  if (length(ethnicity_columns) == 0) {
+standard_filter_headers <- function(meta) {
+  search_strings <- harmonised_col_names %>%
+    pull(col_name_search_string) %>%
```
cjrace, line 2562: This wasn't the way I was expecting you to do it - my first assumption was it would work like the [geographic_catch()](https://github.com/dfe-analytical-services/dfe-published-data-qa/blob/a9570b1248b64f6eb52c3be6f51e92f6e754be60/R/mainTests.r#L2008) with a [regex for key phrases in columns](https://github.com/dfe-analytical-services/dfe-published-data-qa/blob/a9570b1248b64f6eb52c3be6f51e92f6e754be60/R/knownVariables.r#L113)

What you have seems to work fine for the list of aliases we have now though!

cjrace (changes requested):

data/harmonised_col_names.csv
rmbielby: I figured creating an error anytime someone uses just "sen" in a col_name seemed a little risky. Can't think of any obvious times that might happen and not be referring to SEN, but it's 3 of the most used letters in the alphabet, I feel like there's going to be times it comes up where we don't want to flag an error. And all the cases that Louis found in the existing data sets that we'd want to flag contained "sen_" as a substring.

tests/testthat/mainTests/ethnicity_meets_standards.csv
rmbielby: Yes, it could... I've abridged it now.

global.R
rmbielby: Accidental, reverted it now

R/mainTests.r
rmbielby: Sure, switched it over now.

R/mainTests.r
rmbielby: I've made a few small changes now removing the for loop and incorporating filter_grouping_column entries into the check. It's a little more like `geographic_catch()`, but probably still not quite what you were expecting!

data/harmonised_col_names.csv
cjrace: That's fair!

cjrace (approved):

PR #134, opened 2024-09-05
Add numberstring as an acceptable indicator_unit

cjrace:
# Brief overview of changes

Adding numberstring as an acceptable indicator_unit.

We'd added it into EES this week, so the screener should match to allow users to use it.

Added numberstring as an acceptable value
Added new test data and test that contains all indicator units to make sure they pass
Minimised the existing test data for indicator units to reduce size

Links to https://github.com/dfe-analytical-services/analysts-guide/pull/83 as a complementary guidance change, which explains the behaviour.

None


cjrace:
[Update data screener to allow numberstring](https://trello.com/c/QLGnMufW/1719-update-data-screener-to-allow-numberstring)

rmbielby (dismissed):

Issue #135, opened 2024-09-17
Add tracking for files screened


cjrace:
We should set up a database table that we populate with the following columns:

- environment - e.g. 'local', `development`, 'production', `shinyapps.io` (can set environment variables to control)
- data_filename - pulled from the csv data file
- file_size - we should just track the raw bytes so it's easier to analyse, rather than the `dfeR::pretty_filesize()` output
- rows_ count - same thing we put in the UI 
- cols_count - same thing we put in the UI 
- stage - what stage of the checks the file got to
- pass - boolean, did it pass
- warnings - string of the checks that had a warning (take the name from the test col of the screening output table, e.g. "ethnicity_values, total, ob_unit_meta" (we can then parse this if we want to splitting by commas if we want to analyse what warnings are happening a lot)
- time_started - time recorded at start of screening
- time_ended - time recorded at end of screening
- screening_time - count of time taken to screen in raw seconds calculated from above two cols, I think we'd want a raw value like seconds that we can ees-ily average / aggregate

We then add code into the server side of the app that will write a new row into the database table after each screening.

Example UI output now
![image](https://github.com/user-attachments/assets/e55650cf-f15f-4506-b5cd-9d28975b2dfc)

Example screening output table (where we can use the test column to pull from and populate with what warnings are present (as will be interesting thinking of API standards), also shows the 'stages' we have
![image](https://github.com/user-attachments/assets/5acc595f-e299-4180-924d-7d05675e53e2)

### Easy-ish first tasks
- [x] Add time tracking into the screener as is (log time at start of screening, log time at end, present in UI with `dfeR::pretty_time_taken()`
- [ ] Check environment variables exist
- [ ] Check all data above exists in app

### Main tasks
- [ ] Create new database table
- [ ] Add a connection into the app to point to our existing SQL databases (can migrate to databricks at a later point)
- [ ] Add code into server file that gathers all of this and writes a new row into the database

PR #141, opened 2024-09-18
Added screening duration to summary text

rmbielby:
# Brief overview of changes

Adding in a screening duration to the test summary stats. 

This is being added so that it can be included as we introduce metric logging for individual tests.

I've stored the screener trigger time (i.e. when the use triggers the screening to start) and the screening end time as a reactive values instance. So these should now be nicely retrievable however we write out to the summary lines.

I've also added the screener run duration to the summary stats.

Contributes to #135 (is there an appropriate keyword for that?)

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -169,14 +169,7 @@
         "Date/Publication": "2024-06-15 19:40:02 UTC",
         "Encoding": "UTF-8",
         "Built": "R 4.4.0; x86_64-w64-mingw32; 2024-06-16 04:59:25 UTC; windows",
-        "Archs": "x64",
-        "RemoteType": "standard",
-        "RemotePkgRef": "PKI",
-        "RemoteRef": "PKI",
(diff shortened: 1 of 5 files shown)


cjrace (dismissed):
Approving as it's safe to go in. Though I'm not sure on the purpose of the new test files?

Also, don't think there is a keyword for 'contributes' though just referencing it will add a mention on the issue itself so there's an easy link between them automatically.

rmbielby:
> Approving as it's safe to go in. Though I'm not sure on the purpose of the new test files?
> 
> Also, don't think there is a keyword for 'contributes' though just referencing it will add a mention on the issue itself so there's an easy link between them automatically.

Those files look like they're from me demoing a potential tidy data format to someone a while back. I must have had them hanging around in my repo when I created this branch. I've removed them now to keep things tidy.

cjrace (approved):

PR #145, opened 2025-01-02
Link UI to max filesize limit

cjrace:
# Brief overview of changes

Made the UI note about the max file size limit pull from the actual limit set in `global.R`.

Noticed the two were out of sync and probably have been for years, figured it was a quick enough change to make while I was there anyway.

- Set a variable for the limit that is used in `ui.R` and `global.R`
- Fixed the bytes to MB conversion (I'd done it wrong before)
- Used dfeR to pretty the raw bytes properly
- Bumped the value to 900mb, just because I think it'll be fine and we might have some files in the 800mb bracket
- Added a projectId, well R Studio did when I opened it

Project ID thing seems to be standard now so having one will stop R Studio trying to make it each time we open it up.

No related ticket.

### DIFF
--- a/dfe-published-data-qa.Rproj
+++ b/dfe-published-data-qa.Rproj
@@ -1,4 +1,5 @@
 Version: 1.0
+ProjectId: eec6f1e6-a03a-410a-b16d-2f41ff547e70
 
 RestoreWorkspace: Default
 SaveWorkspace: Default
--- a/global.R
+++ b/global.R
@@ -1,5 +1,10 @@
 # Max file size ---------------------------------------------------------------------------------
-options(shiny.maxRequestSize = 800 * 1024^2)
+# Set the maximum size you want in megabytes, then this will convert it as needed to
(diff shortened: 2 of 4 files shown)


rmbielby (approved):
All works fine. Couple of comments:
- I'd never actually noticed that max size statement and I guess no-one had in a while given how out of date it was. I'd be tempted to put it some sort of info / call-out box to highlight it.
- Similarly with the line below it around giving feedback, feels like that's the sort of thing we'd put in a beta banner style element usually.

cjrace:
That's fair, it's definitely not how I'd lay it out now, I think it was literally 'my first UI'. Done a quick update so that info now sits in a panel that persists through to the screening results too, how's this? Suspect with the theme updates to the analysts guide being intended as a potential starting point for a new consistent theme across our apps that the overall styling of this app might be getting a more thorough update in the next few months anyway

image

image

rmbielby (approved):
Nice

PR #147, opened 2025-01-22
Fix filter hint blanks bug

cjrace:
# Brief overview of changes

Added extra filtering into the `filter_hint()` test to fix an issue we had with a file failing despite seeming to be okay. Seemed to be a specific scenario where the CSV values where all quoted, and there was some things going on around encoding.

Gone for the quick fix rather than in depth answering of potential questions around this as I think that's the best use of time.

If we don't fix it the app can fail in a confusing way, which is unhelpful and takes up time for analysts and also whoever is on the mailbox that day.

This is the error and message we got:

> I am getting the following error message but can’t see that there is an indicator with a filter hint in my metadata file (attached). Tried removing any whitespace but still no luck.

image

Started off with a drastically cut down, dummy version of the data set we found the issue with and added an edge case test to track it.
image

Then used the debugging script I read that file in, the `filter_hint()` function that was breaking was using the main metafile version (not the forced character one), and that looks like this:

image

Realising at this point now I'm looking at the code that there's a simpler solution than I thought, it's just failing to filter out the blanks.

image

As I'm sure we've done this elsewhere and just added an extra condition, I've done that. I'm fairly certain EES will handle it fine so this change should help the file to pass the screener okay.

The short term fix for them was resaving the file as UTF-8 out of Excel, they had been saving using `write.csv(..., fileEncoding = "UTF-8")` but it didn't seem to work. Suggested trying `read_csv()` in future for their pipeline as that does the encoding by default, though haven't tested that assumption. 

Though at this point (after the PR is merged), the original code would still work and pass the screener anyway. The details of the encoding changes, and why Excel saved it in a way that was read by the screener as NA, when before it was read as "" is probably something for another day.

No related issue.


cjrace:
[Look into screener bug (UTF-8 meta)](https://trello.com/c/aEZvTM33/2109-look-into-screener-bug-utf-8-meta)

cjrace:
[Look into screener bug (UTF-8 meta)](https://trello.com/c/aEZvTM33/2109-look-into-screener-bug-utf-8-meta)

R/mainTests.r
```diff
@@ -2062,7 +2062,7 @@ geographic_catch <- function(meta) {
 
 filter_hint <- function(meta) {
   filter_hints <- meta %>%
-    filter(col_type == "Indicator", !is.na(filter_hint)) %>%
+    filter(col_type == "Indicator", !is.na(filter_hint), filter_hint != "") %>%
```
rmbielby, line 2065: Old typo in the pass message on line 2075: "have an filter_hint"

tests/testthat/otherData/filter_hint_indicator_bug.csv
rmbielby, line 1: Any particular reason for these test files being in otherData/ rather than mainTests/, given that the test itself is in maintests?

rmbielby (changes requested):
Super minor comments. Only other thing is if it's worth making the existing fail message more informative whilst you're editing this bit of the code? I tested it out with "   " in place of "" and got a fail (which I guess is the behaviour we want?), but no information on how to find the fail. Could we paste the col_names of the indicators with filter_hints in the fail message and maybe the offending values in quatoation marks?

tests/testthat/otherData/filter_hint_indicator_bug.csv
cjrace: All the files in the mainTests/ folder have exact name matches with the function, so there's 1 example file for every function that the test helper function `testIndividualTest()` uses, and then everything else is edge cases on top where you can specify the name of the test data using `testOther()` (or least that's the pattern I've followed to date, suspect we'll want to revise the testing approach / add more specific function tests when moving to a package)

cjrace:
I debated adding the values into the message first time around - it would have saved time in spotting what the issue was. Have added that in now and fixed the typo.

image

rmbielby (approved):

PR #151, opened 2025-04-22
Added Unknown and Not applicable as across the board standard geographies

rmbielby:
# Brief overview of changes

Geographies using z had gotten a bit patchy between different geographies, so I've added a set up to just include **z / Unknown** and **z / Not applicable** across geographies.

This may well help for shifting across to dfeR as that only supplies the ONS standards and not our DfE z codes used by publications.

Help with standardisation.

- I've added `expected_standard_geog_combinations` in **knownVariables.r** and then pulled that into the geography combinations tests in **mainTests.r**.

...

...


cjrace (changes requested):
Is it worth going even further with the `expected_standard_geog_combinations` to include more, like the outside of england ones? Basically anything that would be in more than one lookup probably is universal enough to be in all?

These four lookups all seem to have the same 'special' locations at the end
image

Not hugely enjoying what I'm assuming is all of the Air generated reformatting changes - creates a lot of noise in the file changes! I'd missed the `expected_standard_geog_combinations` line on my first review and only found it after re-reading the PR description

rmbielby:
Yeah, the Air formatter is a bit of a shift to the code with it's handling of line lengths. Almost went back and undid it before raising the PR. I've still left it as is for these files and at least that's those ones done now.

I've shifted now to putting the z geography entries in their own specific look-up file so that it's easier to point analysts to them, and added the Outside of X entries to that look-up.

cjrace (approved):

PR #154, opened 2025-06-30
Added filter_default as optional meta column

rmbielby:
# Brief overview of changes

EES has been accepting a custom filter_default column since Feb 2025. This brings the screener into line with that, such that it doesn't now flag an error if an analyst uploads a meta file with fitler_default as a column.

Consistency with the live EES service

I've kept it simple. Just:
- added a new known variable called optional_meta_cols, which contains the new column name
- added optional_meta_cols to the invalid_meta_cols test in preCheck1.r

### DIFF
--- a/R/knownVariables.r
+++ b/R/knownVariables.r
@@ -20,6 +20,10 @@ mandatory_meta_cols <- c(
   "filter_grouping_column"
 )
 
+optional_meta_cols <- c(
+  "filter_default"
+)
+
 # Acceptable time identifiers =========================================================================================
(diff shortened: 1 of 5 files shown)


cjrace (approved):
The minimal version works smoother than I'd expected actually!

We definitely need to update the total test at some point soon, and update the analysts guide too / promote more widely (next months top tip maybe?). Though given we're moving across to the new package soon anyway, happy to let this go in as it is to stop the filter_default being an error for teams in the short term

PR #155, opened 2025-07-01
adding sen unclassified to sen provision

mzayeddfe:
# Brief overview of changes

Added sen unclassified as an option for sen_provision

to expand sen_provision categories for ks2.

should I change it to sen unknown instead? phoebe mentioned they're moving towards that label in that email she sent to both of us. however, i noticed we used unclassified in our labels rather than unknown?


mzayeddfe:
added all sen provision for sen_provision

rmbielby (approved):

PR #156, opened 2025-07-16
Updating ethnicity minor categories

mzayeddfe:
# Brief overview of changes

Added ethnicity minor categories that are needed by the KS2 team.

They publish a label we have as two separate ones and not as a single label.
We're also missing the All option for the mixed label. 

Added the following categories:

ethnicity_minor - Gypsy
ethnicity_minor - Irish Traveller
ethnicity_minor - All Mixed / Multiple ethnic groups


mzayeddfe:
added total labels for ethnicity_minor

rmbielby (approved):
Looks good, thanks for the extra bits around the edges

PR #158, opened 2025-07-25
Adding apprenticeship categories to data dictionary

rmbielby:
# Brief overview of changes

Adding some categories relating to apprenticeship data that's going out via the API.

I've also added a contexts column to highlight any specific areas that a given item is used in.


mzayeddfe (changes requested):
This is an interesting PR because I haven't interacted much with apprenticeship data!

I have a couple of things I wanted to check:

1.  Should we provide more context for the 'Unknown' label in provider_type or does this not apply here? Something like 'Unknown provider type'
2. Should we also make the 'Other' label for funding_type more clear like 'Other type of funding' or does this not apply here?
3. We have achievement_count and achievements_percent, should achievement be single or plural for both?

Let me know what you think! :) 

NOTE: i know you said one of the workflow wasn't working this morning but i wanted to flag just in case. it's the automated tests one.

rmbielby:
> This is an interesting PR because I haven't interacted much with apprenticeship data!
> 
> I have a couple of things I wanted to check:
> 
> 1. Should we provide more context for the 'Unknown' label in provider_type or does this not apply here? Something like 'Unknown provider type'
> 2. Should we also make the 'Other' label for funding_type more clear like 'Other type of funding' or does this not apply here?
> 3. We have achievement_count and achievements_percent, should achievement be single or plural for both?
> 
> Let me know what you think! :)
> 
> NOTE: i know you said one of the workflow wasn't working this morning but i wanted to flag just in case. it's the automated tests one.

For 1 and 2, I think there's enough context in the overall filter label that we don't need to worry about any extra text. "Unknown" especially is consistent with other geographies and also how we do ethnicity.

For the plural consistency, that would be neater, but I think we can let it pass as this is what the team have gone for.

mzayeddfe (approved):

PR #160, opened 2025-07-31
Filter grouping columns have own row in meta

### LINKED ISSUE #159: [Feature]: Remove error on filter grouping column being included as a row in the meta
### Description

For the API to include filter grouping columns in the API, then the filter grouping column needs its own row in the meta data file. Currently the screener gives an error if this is the case. What we now need is a warning if the filter grouping column does *not* have a row in the meta.

### Use case

I want to be able to query on filter grouping columns in the API

### Why this is important

So that all required fields are available via the API

### Additional Context

Good example is the apprenticeships data

rmbielby:
# Brief overview of changes

For the API to have access to a filter_grouping_column, it needs it's own row in the meta file for a data set. Previously, this has raised an error in the screener and, if uploaded to EES, shown the filter_grouping_column as a filter option as well as a grouping.

This has been switched on EES such that the table tool now only uses a filter grouping column as a group and doesn't create a set of selectable options.

This means the data file requirements have changed so that:
- instead of being a FAIL if a filter_grouping_column has its own row, it should be a pass
- instead of being a PASS if a filter_grouping_column doesn't have its own row, it should be a WARNING

To fit current EES table tool functionality and allow the API to return filter grouping columns.

I've changed the test filter_group_not_filter to filter_group_is_filter

Closes #159


R/mainTests.r
```diff
@@ -2886,13 +2886,14 @@ filter_group_level <- function(data, meta) {
   return(output)
 }
 
-# filter_group_not_filter -------------------------------------
+# filter_group_is_filter -------------------------------------
 # Checking that filter groups are not filters
 
-filter_group_not_filter <- function(meta) {
+filter_group_is_filter <- function(meta) {
+  standard_cols_allowed <- c("school_name")
```
cjrace: Trying to follow the logic on `school_name` being a special case and can't work it out, I blame the sleep deprivation - what's your thinking on this?

cjrace (commented):
Can't lie, the flip from 'is not' to 'is' really hurt my head when reviewing! All checks out though except for one minor question inline.

I'm assuming you've thought this through, though wanted to highlight this change isn't backwards compatible. I think that's fine as it mostly affects API data sets, which is forward looking only, but thought I'd flag it's effectively a major breaking change to our data standards.

1. This will relax current checks that prevent issues from including the rows in the metadata, so is contingent on https://dfedigital.atlassian.net/browse/EES-5840 being completed, maybe safest to avoid merging until that's done?

2. We need to update the analysts guide as well as our current guidance there is very explicitly the opposite of the new behaviour (added an issue for this - https://github.com/dfe-analytical-services/analysts-guide/issues/220)

R/mainTests.r
rmbielby: Hah, I'm not sure my logic would have been discernible, even without sleep deprivation! It's probably best to just remove this to be honest. From what I remember, the reason it ended up being included was:
- I picked a random file as a test file for this work
- It had school name as a column and some other filter, so I just set school name to be the filter_grouping_column and added school name to the meta file
- The screener gave an error because school name is a protected column name and it shouldn't be included in the meta file
- Rather than returning the meta file back to it's original form, I thought "what if someone did actually want to have a child-filter of school name or some other protected column name?" and added a workaround to allow it.

Removing it now as this doesn't sound like something we need (or that's been tested on EES in terms of how it would actually behave).

rmbielby:
> I'm assuming you've thought this through, though wanted to highlight this change isn't backwards compatible. I think that's fine as it mostly affects API data sets, which is forward looking only, but thought I'd flag it's effectively a major breaking change to our data standards.
> 
> 1. This will relax current checks that prevent issues from including the rows in the metadata, so is contingent on https://dfedigital.atlassian.net/browse/EES-5840 being completed, maybe safest to avoid merging until that's done?

Yeah, that was why I kept it as a warning rather than an error. So analysts can still stick to the original convention, but nudged towards the new one.

R/mainTests.r
cjrace: I do feel better now - no way I was figuring that one out!

cjrace (approved):
Happy to approve - there's some conflicts but take this as approval to fix those and merge (or nudge me to reapprove if you can't merge after that)

We should priotitise https://github.com/dfe-analytical-services/analysts-guide/issues/220 now this is in and shout about it on Teams and in the next show and tell

PR #162, opened 2025-10-07
Data dictionary/key stage 4

rmbielby:
# Brief overview of changes

Adding Key Stage 4 elements to the data dictionary as agreed with the Key Stage 4 team.

This also adds some broader elements (i.e. used by KS4, but that will also be widely used across other publications).

This includes some elements on SEN, ethnicity etc and uses the breakdown_topic / breakdown combo in places.

As well as adding in the new breakdowns, I've got some updates to the data checker code and some tests added on for validating the data dictionary.


data/data-dictionary.csv
```diff
@@ -132,62 +184,274 @@ establishment_type_group,Filter,Academies and free schools,,,,api-only,Establish
...
+grade,Filter,33,,,,api-only,Attainnment
+grade,Filter,43,qualification_detailed,GCSE (9-1) Full Course (Double Award),,api-only,Attainnment
+grade,Filter,43,,,,api-only,Attainnment
+grade,Filter,44,qualification_detailed,GCSE (9-1) Full Course (Double Award),,api-only,Attainnment
+grade,Filter,44,,,,api-only,Attainnment
+grade,Filter,54,qualification_detailed,GCSE (9-1) Full Course (Double Award),,api-only,Attainnment
+grade,Filter,54,,,,api-only,Attainnment
+grade,Filter,55,qualification_detailed,GCSE (9-1) Full Course (Double Award),,api-only,Attainnment
+grade,Filter,55,,,,api-only,Attainnment
+grade,Filter,65,qualification_detailed,GCSE (9-1) Full Course (Double Award),,api-only,Attainnment
```
cjrace: Is it worth just adding all numbers from 1-100 in this group for future?

data/data-dictionary.csv
```diff
@@ -226,19 +494,44 @@ provider_type,Filter,Special college,,,,api-only,Establishment characteristics
...
+school_religious_character,Filter,Hindu,,,,api-only,Establishment characteristics
 school_religious_character,Filter,Jewish,,,,api-only,Establishment characteristics
 school_religious_character,Filter,Methodist,,,,api-only,Establishment characteristics
+school_religious_character,Filter,Multi-faith,,,,api-only,Establishment characteristics
 school_religious_character,Filter,Muslim,,,,api-only,Establishment characteristics
 school_religious_character,Filter,No religious character,,,,api-only,Establishment characteristics
 school_religious_character,Filter,Other Christian faith,,,,api-only,Establishment characteristics
 school_religious_character,Filter,Other religious character,,,,api-only,Establishment characteristics
+school_religious_character,Filter,Roman catholic,,,,api-only,Establishment characteristics
 school_religious_character,Filter,Roman Catholic,,,,api-only,Establishment characteristics
```
cjrace: I don't think we should have both versions - quick google suggests it should be capitalised 'Catholic'?

data/data-dictionary.csv
```diff
@@ -284,49 +578,353 @@ sen_secondary_need,Filter,Specific learning difficulty,,,SPLD,api-only,Special e
...
+subject,Filter,Any classical study,,,,api-only,Attainment
+subject,Filter,Any design & technology,,,,api-only,Attainment
+subject,Filter,Any English,,,,api-only,Attainment
+subject,Filter,Any mathematics,,,,api-only,Attainment
+subject,Filter,Any modern language,,,,api-only,Attainment
+subject,Filter,Any Science,,,,api-only,Attainment
+subject,Filter,Any Subject,,,,api-only,Attainment
+subject,Filter,Arabic,,,,api-only,Attainment
+subject,Filter,Art & Design,,,,api-only,Attainment
+subject,Filter,Art and Design,,,,api-only,Attainment
```
cjrace: We should probably only have the 'and' version of these two?

data/data-dictionary.csv
```diff
@@ -284,49 +578,353 @@ sen_secondary_need,Filter,Specific learning difficulty,,,SPLD,api-only,Special e
...
+sen_status,Filter,Total,,,,api-only,Special educational needs
 sex,Filter,Boys,,,,api-only,Personal characteristics
 sex,Filter,Girls,,,,api-only,Personal characteristics
 sex,Filter,Other,,,,api-only,Personal characteristics
 sex,Filter,Total,,,,api-only,Personal characteristics
+subject,Filter,All Subjects,,,,api-only,Attainment
+subject,Filter,Ancient History,,,,api-only,Attainment
+subject,Filter,Animal Care,,,,api-only,Attainment
+subject,Filter,Any classical study,,,,api-only,Attainment
+subject,Filter,Any design & technology,,,,api-only,Attainment
```
cjrace: 'and' instead of &?

data/data-dictionary.csv
```diff
@@ -284,49 +578,353 @@ sen_secondary_need,Filter,Specific learning difficulty,,,SPLD,api-only,Special e
...
+subject,Filter,Childcare,,,,api-only,Attainment
+subject,Filter,Chinese,,,,api-only,Attainment
+subject,Filter,Classical Civilisation,,,,api-only,Attainment
+subject,Filter,Classical Greek,,,,api-only,Attainment
+subject,Filter,Combined Science,,,,api-only,Attainment
+subject,Filter,Communication Studies,,,,api-only,Attainment
+subject,Filter,Computer Science,,,,api-only,Attainment
+subject,Filter,Construction,,,,api-only,Attainment
+subject,Filter,Dance,,,,api-only,Attainment
+subject,Filter,Design & Technology,,,,api-only,Attainment
```
cjrace: 'and' instead of '&'

data/data-dictionary.csv
```diff
@@ -284,49 +578,353 @@ sen_secondary_need,Filter,Specific learning difficulty,,,SPLD,api-only,Special e
...
+subject,Filter,Hair / Personal Care,,,,api-only,Attainment
+subject,Filter,Health,,,,api-only,Attainment
+subject,Filter,History,,,,api-only,Attainment
+subject,Filter,Hospitality and Catering,,,,api-only,Attainment
+subject,Filter,Information and Communication Technology,,,,api-only,Attainment
+subject,Filter,Italian,,,,api-only,Attainment
+subject,Filter,Land Based Studies,,,,api-only,Attainment
+subject,Filter,Latin,,,,api-only,Attainment
+subject,Filter,Law,,,,api-only,Attainment
+subject,Filter,Mathematics,,,,api-only,Attainment
```
cjrace, line 744: Is Mathematics any different to Maths?

data/data-dictionary.csv
```diff
@@ -284,49 +578,353 @@ sen_secondary_need,Filter,Specific learning difficulty,,,SPLD,api-only,Special e
...
+subject_discount_group,Filter,Business Studies,subject,Business,,api-only,Attainment
+subject_discount_group,Filter,Chemistry (General),subject,Chemistry,,api-only,Attainment
+subject_discount_group,Filter,Childcare Skills,subject,Childcare,,api-only,Attainment
+subject_discount_group,Filter,Chinese,subject,Other Modern Languages,,api-only,Attainment
+subject_discount_group,Filter,Classical Civilisation,subject,Classical Civilisation,,api-only,Attainment
+subject_discount_group,Filter,Community Development,subject,Social Studies,,api-only,Attainment
+subject_discount_group,Filter,Computer Appreciation / Introduction,subject,Information and Communication Technology,,api-only,Attainment
+subject_discount_group,Filter,Computer Science,subject,Computer Science,,api-only,Attainment
+subject_discount_group,Filter,Construction and the Built Environment,subject,Construction,,api-only,Attainment
+subject_discount_group,Filter,Construction and the Built Environment (Construction),subject,Construction,,api-only,Attainment
```
cjrace: Are these just the same subject?

data/data-dictionary.csv
```diff
@@ -284,49 +578,353 @@ sen_secondary_need,Filter,Specific learning difficulty,,,SPLD,api-only,Special e
...
+subject_discount_group,Filter,Retailing,subject,Business,,api-only,Attainment
+subject_discount_group,Filter,Russian,subject,Other Modern Languages,,api-only,Attainment
+subject_discount_group,Filter,Science Double Award,subject,Combined Science,,api-only,Attainment
+subject_discount_group,Filter,Sociology,subject,Social Studies,,api-only,Attainment
+subject_discount_group,Filter,Spanish,subject,Spanish,,api-only,Attainment
+subject_discount_group,Filter,Speech & Drama,subject,Drama,,api-only,Attainment
+subject_discount_group,Filter,Speech & Drama,subject,Speech & Drama,,api-only,Attainment
+subject_discount_group,Filter,Sports / Movement Science,subject,Sports,,api-only,Attainment
+subject_discount_group,Filter,Sports Studies,subject,Physical Education,,api-only,Attainment
+subject_discount_group,Filter,Sports Studies,subject,Sports,,api-only,Attainment
```
cjrace, line 845: Should these be standardised to one row?

R/manual_scripts/api-data-checker.r
cjrace, line 1: Only just realised, should this script be in the /manual_scripts folder?

cjrace (changes requested):
I think we need to start thinking about automated tests on the data dictionary to spot repeated rows and check for special symbols we want to avoid like '&', which we should swap for 'and' etc?

Left a few comments on rows that look like we probably should be standardising

data/data-dictionary.csv
rmbielby: I believe there are only specific combinations that are achievable here. 1 to 9 and then some double combinations. So 76 isn't seventy-six, it's their way of presenting double award in Science I think.

data/data-dictionary.csv
rmbielby: Removed the extra one

data/data-dictionary.csv
rmbielby: Done

data/data-dictionary.csv
rmbielby: Done

data/data-dictionary.csv
rmbielby: KS2 used Maths. KS4 used Mathematics...

data/data-dictionary.csv
rmbielby: I've double checked their data and one's used for one qualification type and the other for some other types of qualification. There's no overlap, so looks like they should just be normalised to the same thing.

data/data-dictionary.csv
rmbielby: I think they're genuinely grouped differently here depending on what qualification is being done. GCSEs Sports Studies is classed under Physical Education and the vocational qualifications group it under Sports. I think they do want to make the distinction.

R/manual_scripts/api-data-checker.r
rmbielby: Good idea.

data/data-dictionary.csv
rmbielby: Removed the ampersand version

cjrace (approved):
Thanks @rmbielby!

PR #163, opened 2025-10-07
Data dictionary/key stage 2

rmbielby:
# Brief overview of changes

Adding Key Stage 2 data items into the data dictionary. These are items as used in the Key stage 2 attainment publication data released in September 2025.

Standardising variables based on Key Stage 2 API data sets

The data dictionary has been updated with relevant items. I've also created a quick stand alone function to output fields that aren't consistent with the data dictionary. This is intended to work on multiple input files, so that you can collate all items across a single publication's data sets.

Some package versions that still need fixing for deploy to go through:
- curl@5.2.3
- httr2@1.0.6
- openssl@2.2.2

This PR will be followed up by Phonics and Key stage 4 as well, which will have some iterations on the API data checker function.


data/data-dictionary.csv
cjrace, line 1: Why is row 260 (Total) empty for contexts and not special educational needs?

Image

cjrace: Similar for 246, which isn't a total

Image

cjrace: Should ethnicity be personal characteristics? There's a few parts of that column that seem odd they're empty?

R/api-data-checker.r
```diff
@@ -0,0 +1,97 @@
+# This is a standalone function not run by the screener itself.
```
cjrace, line 1: Could you give an example of using the function / what format to feed files in with? (or move the checker example run to the top of the script?)

global.R
```diff
@@ -2,7 +2,7 @@
 # Set the maximum size you want in megabytes, then this will convert it as needed to
 # - Show the note in the UI 'notes' section
 # - Set the max file upload size for the app
-max_file_size_mb <- 900
+max_file_size_mb <- 2000
```
cjrace, line 5: Nice!

R/api-data-checker.r
```diff
@@ -0,0 +1,97 @@
...
+  entries <- data.frame(
+    col_name = NA,
+    col_type = NA,
+    filter_item = NA,
+    col_name_parent = NA,
+    filter_item_parent = NA
+  )
+  for (file in files) {
+    data <- vroom::vroom(file)
+    meta <- vroom::vroom(gsub(".csv", ".meta.csv", file))
```
cjrace: Might be nice to reduce the console outputs?
```suggestion
    data <- vroom::vroom(file, show_col_types = FALSE)
    meta <- vroom::vroom(gsub(".csv", ".meta.csv", file), show_col_types = FALSE)
```

R/api-data-checker.r
```diff
@@ -0,0 +1,97 @@
...
+        col_name_parent,
+        filter_item_parent,
+        filter_item
+      ) |>
+      dplyr::distinct()
+  )
+}
+
+non_dd_rows <- function(listing) {
+  dd <- vroom::vroom("data/data-dictionary.csv") |>
```
cjrace:
```suggestion
  dd <- vroom::vroom("data/data-dictionary.csv", show_col_types = FALSE) |>
```

R/api-data-checker.r
```diff
@@ -0,0 +1,97 @@
...
+      col_name_parent,
+      filter_item_parent
+    ) |>
+    dplyr::mutate(across(everything(), ~ dplyr::if_else(is.na(.x), "", .x)))
+  listing |>
+    dplyr::anti_join(dd)
+}
+
+checker_example_run <- function() {
+  dir <- "../../offline-data/ks2_attainment/2025-september/Publication\ files/"
```
cjrace: Could leave as a path within the package itself so it's more reproducible?
```suggestion
  dir <- "tests/testthat/test-data/"
```

Though that did fall over for me, might be a better alternative

Image

global.R
cjrace, line 5: Nice!

cjrace (changes requested):
Left a couple of comments. Main ones are around the dictionary itself, for the checker script, I'm happy with that if you are, but made a couple of suggestions to make it a bit more user friendly if you want to

data/data-dictionary.csv
rmbielby: Fair, I've just done a review of that column and entered something for everything now. I'm mainly thinking of it as a useful reference guide for our side so we can see where a given entry roughly originates from.

R/api-data-checker.r
rmbielby: Took the simpler option...

cjrace: Fair, will be good to package / function up more properly still but that can wait for eesyscreener

cjrace (approved):

PR #164, opened 2025-10-13
Adding new Sheffield / Barnsley LA codes

rmbielby:
# Brief overview of changes

Adding in new Sheffield / Barnsley codes.

Analysts need to be able to use new LA boundaries created in May 2025.

New new_la_codes for Barnsley and Sheffield. GIAS haven't updated their old_la_codes at all yet, but notice that for Gateshead they just kept the original old_la_code anyway despite the new_la_code changing. Maybe to reduce impact on schools / annoying them with minor code changes. Will need to keep an eye out for how they respond with these ones.

...

...


cjrace (approved):
Happy to approve, seems logical to keep the assumption that the codes won't change until we hear otherwise.

Not sure why the region_la_hierarchy has been stripped of quotes, doesn't make a material difference but odd to happen?

rmbielby:
> Happy to approve, seems logical to keep the assumption that the codes won't change until we hear otherwise.
> 
> Not sure why the region_la_hierarchy has been stripped of quotes, doesn't make a material difference but odd to happen?

Seems to have done it selectively for just ones where it's needed now.

PR #167, opened 2025-11-14
Small changes to provider types for the apprenticeships data

rmbielby:
# Brief overview of changes

Apprenticeships are just adjusting their provider types a little for an upcoming amendment. This adapts the data dictionary to meet their current provider descritpions.

Keep the data dictionary in line with the apprenticeship data.

Local authority and Other are categories that have been added to their data and I've just adjusted existing entries (that weren't actually in use yet in live data sets) to match.

They're also discontinuing publication of "Unknown" and "Other public funded...", but I'm leaving these in in case there becomes a need for them again (they were removed based on consultation with the sector saying that they weren't interested in those breakdowns).

...


cjrace (approved):
Looks good, and happy to leave in the other options that the sector aren't interested in right now, they seem like valid entries if they were ever used

PR #168, opened 2025-11-14
Geography update: Adding Sheffield and Barnsley to the LADs list

rmbielby:
# Brief overview of changes

I'd previously added the new codes for Sheffield and Barnsley to the LA lookups, but not the LAD one. This PR adds those new codes in. I've also added some tests to double check that the number of E06 and E08 authorities match across the two files as both of these count as both LAs and LADs.

Data consistency.

Spotted this in relation to a query from @robert-tarrant

Also ordered both files by the ONS codes as neither seemed to have a clear and intuitive ordering and it was bugging me (sorry that makes the actual data additions a lot less obvious in the diffs). The actual meaningful changes are on lines 295, 298, 317 and 318 of the updated lads.csv.


tests/testthat/test-data-geographies.R
robert-tarrant, line 1: Would the filtering at the start make the E08 check invalid as all E08 LA / LADs are filtered out? When running manually the outputs both return 0, where they should be 39 (i.e. it'd be possible for the check to pass even if the count didn't match)

rmbielby: Ah yes! Good spot. Meant to delete that filtering after moving the code around. Done that now.

robert-tarrant (approved):
Looks good!

PR #171, opened 2025-11-18
Added police force areas as a geographic level option

rmbielby:
# Brief overview of changes

Adding Police force area to geographic levels.

Some teams need to report stats on police force areas

I've added Police force area as a geographic level and added the codes and names straight from ONS. Example has been added to the passes everything test file.


cjrace (approved):
Have raised https://github.com/dfe-analytical-services/dfeR/issues/129 as ideally we'd document getting the data more, and do it in a reproducible way in dfeR instead (particularly thinking of when we move to eesyscreener and need to reliably access through code rather than a CSV).

Otherwise, it all works, all seems sensible, so happy to approve!

PR #177, opened 2026-01-16
Added CME and EHE filters and indicators to Data Dictionary

CHall2342:
# Brief overview of changes

Added filters and indicators relevant to the CME and EHE data.

So that CME and EHE data can be brought into the API.

For CME: I've added different labels for the Sex breakdown_topic, the Year Group, Duration and Reason filters as breakdowns and the three indicator columns used in the CME data (child_count, child_percent, rate_per_100) to data_dictionary.csv.

For EHE: I've added the ehe_status groupings for prior_settings, as well as three indicators related to notices/orders being issued or revoked.

### DIFF
--- a/data/data-dictionary.csv
+++ b/data/data-dictionary.csv
@@ -153,8 +153,6 @@ breakdown,Filter,Vision impairment,breakdown_topic,SEN primary need,VI,api-only,
 breakdown,Filter,White and Asian,breakdown_topic,Ethnicity,,api-only,Personal characteristics - Ethnicity
 breakdown,Filter,White and Black African,breakdown_topic,Ethnicity,,api-only,Personal characteristics - Ethnicity
 breakdown,Filter,White and Black Caribbean,breakdown_topic,Ethnicity,,api-only,Personal characteristics - Ethnicity
-breakdown,Filter,Male,breakdown_topic,Sex,,api-only,Personal characteristics
-breakdown,Filter,Female,breakdown_topic,Sex,,api-only,Personal characteristics
 breakdown,Filter,Unknown,breakdown_topic,Sex,,api-only,Personal characteristics
 breakdown,Filter,Reception,breakdown_topic,Year group,,api-only,Personal characteristics
 breakdown,Filter,Year 1,breakdown_topic,Year group,,api-only,Personal characteristics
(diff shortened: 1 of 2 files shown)


data/data-dictionary.csv
```diff
@@ -150,6 +150,51 @@ breakdown,Filter,Unknown,breakdown_topic,Free school meal eligibility,,api-only,
 breakdown,Filter,Unknown,breakdown_topic,SEN primary need,,api-only,Special educational needs
 breakdown,Filter,Unknown,breakdown_topic,SEN provision,,api-only,Personal characteristics
 breakdown,Filter,Vision impairment,breakdown_topic,SEN primary need,VI,api-only,Special educational needs
+breakdown,Filter,White and Asian,breakdown_topic,Ethnicity,,api-only,Personal characteristics - Ethnicity
+breakdown,Filter,White and Black African,breakdown_topic,Ethnicity,,api-only,Personal characteristics - Ethnicity
+breakdown,Filter,White and Black Caribbean,breakdown_topic,Ethnicity,,api-only,Personal characteristics - Ethnicity
```
rmbielby: I think for the Ethnicities, it's going to be much clearer if we get teams to stick to Ethnicity minor and Ethnicity major for consistency across publications. So let's delete these "Ethnicity" rows from the breakdown columns and I can encourage this team to make the switch later down the line.

data/data-dictionary.csv
```diff
@@ -1344,3 +1404,9 @@ working_below_pupil_count,Indicator,,,,,api-only,Attainment - Key stage 2
 working_below_pupil_percent,Indicator,,,,,api-only,Attainment - Key stage 2
 working_towards_pupil_count,Indicator,,,,,api-only,Attainment - Key stage 2
 working_towards_pupil_percent,Indicator,,,,,api-only,Attainment - Key stage 2
+child_count,Indicator,,,,,api-only,General
+child_percent,Indicator,,,,,api-only,General
+rate_per_100,Indicator,,,,,api-only,General
```
rmbielby: Would be good to add the context on this one. I assume it's `child_rate_per_100`?

rmbielby (changes requested):
Handful of small things I'd like to change looking at it fresh. If you're not 100% what the rate_per_100 refers to, then check with the team. For the ethnicity, I think let's just remove from the dictionary edits and we can discuss with the team further down the line. Would rather they just use Ethnicity minor and Ethnicity major even if it's only one of them that they're reporting.

data/data-dictionary.csv
CHall2342: Think this was a quirk of the order in which branches were made, I didn't add these in - happy to remove these.

data/data-dictionary.csv
CHall2342: Happy to change this

rmbielby (approved):

